PR #28, opened 2020-07-21
finish security policies

janmazak:
`signTx_handleOutput_ui_runStep` is not up-to-date and has to be changed to accomodate outputs given as address params.

### DIFF
--- a/src/bip44.c
+++ b/src/bip44.c
@@ -103,7 +103,7 @@ uint32_t bip44_getChainTypeValue(const bip44_path_t* pathSpec)
 	return pathSpec->path[BIP44_I_CHAIN];
 }
 
-bool bip44_hasValidChainTypeForAddress(const bip44_path_t* pathSpec)
+static bool bip44_hasValidChainTypeForAddress(const bip44_path_t* pathSpec)
 {
 	if (!bip44_containsChainType(pathSpec)) return false;
 	const uint32_t chainType = bip44_getChainTypeValue(pathSpec);
@@ -129,6 +129,12 @@ bool bip44_hasReasonableAddress(const bip44_path_t* pathSpec)
 	if (!bip44_containsAddress(pathSpec)) return false;
 	const uint32_t address = bip44_getAddressValue(pathSpec);
 	return (address <= MAX_REASONABLE_ADDRESS);
+}
+
+// path is valid as the spending path in all addresses except REWARD
+bool bip44_isValidAddressPath(const bip44_path_t* pathSpec)
+{
+	return bip44_hasValidChainTypeForAddress(pathSpec) && bip44_containsAddress(pathSpec);
 }
 
 // Staking keys (one per account, should end with /2/0 after account)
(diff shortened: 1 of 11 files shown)


src/securityPolicy.c
```diff
@@ -47,10 +65,15 @@ static inline bool has_cardano_prefix_and_any_account(const bip44_path_t* pathSp
...
 {
-	return bip44_hasValidChainTypeForAddress(pathSpec) &&
-	       bip44_containsAddress(pathSpec);
+	if (!bip44_hasValidCardanoPrefix(pathSpec))
+		return false;
+
+	if (bip44_isValidStakingKeyPath(pathSpec))
+		return true;
+
+	return bip44_hasValidChainTypeForAddress(pathSpec) && bip44_containsAddress(pathSpec);
```
ppershing, line 76: we should have `isValidAddressPath`

src/signTx.c
```diff
@@ -395,31 +388,36 @@ static void signTx_handleInputAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t w
 
 // ============================== OUTPUTS ==============================
 
+static void ui_displayAmount(uint64_t amount, ui_callback_fn_t callback)
```
ppershing, line 391: I guess we can pass header here. This way you can probably reuse it also for the fee ...

src/signTx.c
```diff
@@ -434,7 +432,95 @@ static void signTx_handleOutput_ui_runStep()
...
+	ui_callback_fn_t* this_fn = signTx_handleOutput_addressParams_ui_runStep;
+
+	UI_STEP_BEGIN(ctx->ui_step);
+
+	UI_STEP(HANDLE_OUTPUT_ADDRESSPARAMS_STEP_DISPLAY_AMOUNT) {
+		ui_displayAmount(ctx->stageData.output.amount, this_fn);
+	}
+	UI_STEP(HANDLE_OUTPUT_ADDRESSPARAMS_STEP_DISPLAY_SPENDING_PATH) {
+		char pathStr[100];
+		{
```
ppershing, line 505: In the similar style -- don't we want to have ui_displaySpendingPath helper? I actually like the idea of such specialized ui utilities. Maybe we can collect it in one file uiScreens ...

src/securityPolicy.c
```diff
@@ -71,6 +94,7 @@ static inline bool is_too_deep(const bip44_path_t* pathSpec)
 #define PROMPT_IF(expr)    if (expr)    return POLICY_PROMPT_BEFORE_RESPONSE;
 #define ALLOW_IF(expr)     if (expr)    return POLICY_ALLOW_WITHOUT_PROMPT;
 #define SHOW_IF(expr)      if (expr)    return POLICY_SHOW_BEFORE_RESPONSE;
+#define SHOW_UNLESS(expr)  if (!(expr)) return POLICY_SHOW_BEFORE_RESPONSE;
```
ppershing, line 97: I guess we might also add WARN_UNLESS

janmazak:
Larger changes will be in separate PRs.

src/bip44.c
```diff
@@ -103,7 +103,7 @@ uint32_t bip44_getChainTypeValue(const bip44_path_t* pathSpec)
 	return pathSpec->path[BIP44_I_CHAIN];
 }
 
-bool bip44_hasValidChainTypeForAddress(const bip44_path_t* pathSpec)
+static bool bip44_hasValidChainTypeForAddress(const bip44_path_t* pathSpec)
```
ppershing, line 106: why do we need static here?

janmazak: We don't, it's just a rather specific function used only here, so `static` is here to avoid leaking the name outside of this compilation unit. (Many of our .c files consistently use static for this purpose, but I think not all of them.)

PR #29, opened 2020-07-23
ui screens

janmazak:
Add ui screens, repeatedly used.


ppershing:
LGTM. I assume you just moved original implementations and I don't need to verify them

janmazak:
Yes. When there was more than one, I tried to include all the asserts from all of them :)

PR #30, opened 2020-07-23
hex utils

janmazak:
Rename the file and remove various unused #includes.


ppershing:
In general LGTM. But I am not sure you did not remove accidentally more includes than you should have. Rule of thumb -- if I use something from header and I am depending on that header transitively, I still want it as an explicit dependency (transitive dependencies have tendency to break stuff when you change seemingly unrelated code)

PR #33, opened 2020-07-28
Security fixes

ppershing:


ppershing:
that memset is intentional, it is setting 'X' and not zero :-)

PR #34, opened 2020-07-30
Fix epoch/slot calculation after shelley parameter change

ppershing:


src/textUtils.c
```diff
@@ -53,13 +53,35 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
 	return rawSize;
 }
 
+
+// TODO: This is valid only for mainnet
+static struct {
+	uint64_t startBlockNumber;
```
janmazak, line 59: Does this mean absolute slot number? There can be fewer blocks than slots.

src/textUtils.c
```diff
@@ -53,13 +53,35 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
...
+};
+
 size_t str_formatTtl(uint64_t ttl, char* out, size_t outSize)
 {
 	ASSERT(outSize < BUFFER_SIZE_PARANOIA);
 
-	const uint64_t SLOTS_IN_EPOCH = 21600;
-	uint64_t epoch = ttl / SLOTS_IN_EPOCH;
-	uint64_t slotInEpoch = ttl % SLOTS_IN_EPOCH;
+	unsigned i = 0;
```
janmazak, line 71: uint64_t?

ppershing: not needed. we are indexing a rather small array

src/textUtils.c
```diff
@@ -53,13 +53,35 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
...
+	while (ttl < EPOCH_SLOTS_CONFIG[i].startBlockNumber) {
+		i++;
+		ASSERT(i < ARRAY_LEN(EPOCH_SLOTS_CONFIG));
```
janmazak, line 74: Is inability to display TTL enough to crash and not sign a tx?...

src/textUtils.c
ppershing: yep, that is absolute slot number. I did not want to confuse slots (which are usually meant per-epoch) with this absolute number

src/textUtils.c
ppershing: Yes. Plus, this is actually buffer overflow check

PR #36, opened 2020-08-20
Bugfix - FPass reward address instead of staking key hash in withdrawals

refi93:
@PeterBenc noticed that transactions with withdrawals are not being serialized properly in Ledger.

Turns out we were serializing the withdrawals in Ledger wrong. We were passing the staking key hash instead of the reward address. Compare with trezor implementation: https://github.com/trezor/trezor-firmware/blob/8d39ba8314fe9cda420c3cdce071562584eddd2c/core/src/apps/cardano/sign_tx.py#L278

This PR fixes that. The corresponding integration tests are fixed in https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/5


src/signTx.c
```diff
@@ -996,20 +996,25 @@ static void signTx_handleWithdrawalAPDU(uint8_t p2, uint8_t* wireDataBuffer, siz
...
-		// add to tx
-		uint8_t rewardAccount[ADDRESS_KEY_HASH_LENGTH];
+		uint8_t rewardAddress[1 + ADDRESS_KEY_HASH_LENGTH];
 		{
-			VALIDATE(bip44_isValidStakingKeyPath(&ctx->stageData.withdrawal.path), ERR_INVALID_DATA);
-			{
-				write_view_t out = make_write_view(rewardAccount, rewardAccount + SIZEOF(rewardAccount));
-				view_appendPublicKeyHash(&out, &ctx->stageData.withdrawal.path);
-			}
+			addressParams_t rewardAddressParams = (addressParams_t) {
```
janmazak, line 1001: I think that `(addressParams_t)` is superfluous here.

janmazak (approved):

src/txHashBuilder_test.c
```diff
@@ -216,7 +216,7 @@ static struct {
 };
 
 static struct {
-	const char* rewardAccount;
+	const char* rewardAddress;
```
refi93, line 219: worth noting that the original test vector here was 29 bytes, so it was a proper withdrawal with the reward address indeed

PR #38, opened 2020-09-21
Account display format

janmazak:


src/uiScreens.c
```diff
@@ -21,6 +21,60 @@ void ui_displayPathScreen(
 	);
 }
 
+// the given path typically corresponds to an account
+// if it contains anything more, we display just the whole path
```
refi93, line 25: I would expect a condition checking the path length in the code below to leave out printing of the account number, but I don't see anything like that. How/where do you actually check it?

janmazak: I added the missing condition.

src/uiScreens.c
```diff
@@ -21,6 +21,60 @@ void ui_displayPathScreen(
...
+	ASSERT(strlen(screenHeader) > 0);
+	ASSERT(strlen(screenHeader) < BUFFER_SIZE_PARANOIA);
+
+	ASSERT(bip44_hasValidCardanoPrefix(path));
+	ASSERT(bip44_containsAccount(path));
+
+	char accountDescription[160];
+	explicit_bzero(accountDescription, SIZEOF(accountDescription));
+
+	if (bip44_hasReasonableAccount(path) && !bip44_containsChainType(path)) {
```
refi93, line 41: the `bip44_containsChainType` check technically does the check for the path length but it's imho not that obvious for a casual reader of the code. I would rather extend the bip44 library with a function analogous for the `bip44_containsMoreThanAddress` but for Accounts - this would imho make the purpose of the check clear at first sight

src/uiScreens.c
```diff
@@ -21,6 +21,60 @@ void ui_displayPathScreen(
...
+		uint32_t account = bip44_getAccount(path);
+		if (bip44_hasByronPrefix(path)) {
+			snprintf(
+			        accountDescription, SIZEOF(accountDescription),
+			        "Byron account %u  ", account
+			);
+		} else if (bip44_hasShelleyPrefix(path)) {
+			snprintf(
+			        accountDescription, SIZEOF(accountDescription),
+			        "Cardano account %u  ", account
```
refi93, line 51: I tested this with Adalite and it shows me "Cardano account 2147483644 m/1852'/1815'/0'" so it apparently does not handle the hardened index properly

Also, imho the "Cardano" word is stealing too much real estate on the screen and it does not provide much information as the user already willingly chose to open the "Cardano" app (moreover, it can be misleading - Byron is also "Cardano") I would format the message like this instead (ordered by relevance):

`Account #<number> (Shelley/Byron) <path>`

refi93 (changes requested):
1. detected bug regarding handling of hardened paths
2. refactor request for derivation path length check
3. Text reshuffling suggestion

See comments for all three points

refi93:
Also not sure how or when but the current version of the Cardano app in Ledger Live is 2.0.4, not 2.0.1 as set in the current makefile, so it's probably worth bumping to 2.0.5

refi93 (approved):

PR #39, opened 2020-09-30
add support for stake pool registration certificates

janmazak:
There are no restrictions on certificate combinations or other parts of the transaction.


src/textUtils.c
```diff
@@ -105,3 +105,48 @@ size_t str_formatMetadata(const uint8_t* metadataHash, size_t metadataHashSize,
...
+// TODO improve this?
+size_t dnsNameToBuffer(const char* dnsName, uint8_t* buffer, size_t bufferSize)
+{
+	size_t dnsNameLength = strlen(dnsName);
+	ASSERT(dnsNameLength < BUFFER_SIZE_PARANOIA);
+	ASSERT(bufferSize < BUFFER_SIZE_PARANOIA);
+	ASSERT(bufferSize >= dnsNameLength);
+
+	for (size_t i = 0; i < dnsNameLength; i++) {
+		if (dnsName[i] > 127)
```
refi93, line 145: aren't there the same restrictions on character set as for urls?

janmazak: I don't know --- albeit they call it `dns_name` in CDDL, it is not a name, but some type of record, e.g. a space is allowed, probably some forms of newlines, too.

I don't feel it is worth spending too much time on this --- we can't do full validation because it is too complex (and rather annoying to do in C, especially with limited memory), and even a small mistake might lead to denying valid transactions. We don't even display the relays. So I would let the full node decide what is valid and what is not.

We can discuss it at a standup.

ppershing: allow only reasonable part of ascii - something between 32-126 (http://www.asciitable.com/). Basically, just validate that we can at least display it

src/txHashBuilder.c
```diff
@@ -288,20 +291,352 @@ void txHashBuilder_addCertificate_delegation(
...
+		}
+		{
+			BUILDER_APPEND_CBOR(CBOR_TYPE_BYTES, metadataHashSize);
+			BUILDER_APPEND_DATA(metadataHash, metadataHashSize);
+		}
+	}
+	builder->state = TX_HASH_BUILDER_IN_CERTIFICATES;
+}
+
+void txHashBuilder_addPoolRegistrationCertificate_addPoolMetadata_null(
```
refi93, line 610: Pools should have metadata, I don't see why they wouldn't as that's the info by which they are searchable in online tools like pooltool: https://pooltool.io/pool/04c60c78417132a195cbb74975346462410f72612952a7c4ade7e438/blocks

janmazak: I just followed the CDDL spec which allows null, but perhaps we can ask IOHK if we should always require metadata?

janmazak: So I asked if we should support pools with no metadata, and this is what they say: 
"Yes!!! Most of our pools will be private and not have any metadata."

refi93: Thanks for asking and sorry for the confusion, I will update Trezor implementation accordingly

PR #40, opened 2020-10-11
[WIP] Support pool registration certificates

janmazak:
Does not work yet.


src/txHashBuilder.h
```diff
@@ -83,6 +105,44 @@ void txHashBuilder_addCertificate_delegation(
...
+        const uint8_t* rewardAccount, size_t rewardAccountSize,
+        uint16_t numOwners, uint16_t numRelays
+);
+void txHashBuilder_addPoolRegistrationCertificate_enterOwners(tx_hash_builder_t* builder);
+void txHashBuilder_addPoolRegistrationCertificate_addOwner(
+        tx_hash_builder_t* builder,
+        const uint8_t* stakingKeyHash, size_t stakingKeyHashSize
+);
+void txHashBuilder_addPoolRegistrationCertificate_enterRelays(tx_hash_builder_t* builder);
+void txHashBuilder_addPoolRegistrationCertificate_addRelay0(
```
ppershing, line 123: can you explain what is relay0,1,2 and why they are different?

src/securityPolicy.c
```diff
@@ -1,7 +1,9 @@
 #include "addressUtilsShelley.h"
 #include "addressUtilsByron.h"
-#include "securityPolicy.h"
 #include "bip44.h"
+#include "signTx.h"
```
ppershing, line 4: why do we need signTx.h here?

src/securityPolicy.c
```diff
@@ -154,10 +161,15 @@ security_policy_t policyForSignTxInput()
 
 // For each transaction (third-party) address output
 security_policy_t policyForSignTxOutputAddress(
+        bool includePoolRegistrationCertificate,
```
ppershing, line 164: this is bad naming. I think we should use something more like "isSigningPoolRegistrationAsOwner"

src/signTx.c
```diff
@@ -765,12 +799,19 @@ static void signTx_handleCertificate_ui_runStep()
...
+			                        details, SIZEOF(details)
+			                );
+			ASSERT(length == strlen(details));
+			ASSERT(length == 2 * SIZEOF(ctx->stageData.certificate.poolKeyHash));
 			break;
 
+		case CERTIFICATE_TYPE_STAKE_POOL_REGISTRATION:
+			snprintf(title, SIZEOF(title), "Register");
+			snprintf(details, SIZEOF(details), "stake pool");
+			UI_STEP_JUMP(HANDLE_CERTIFICATE_STEP_RESPOND);
```
ppershing, line 813: does this really work this way? I don't remember how exactly was the macro supposed to be used and I don't like we are formatting something without uiDisplay ...

PR #41, opened 2020-10-19
stake pool registration

janmazak:
Some minor things are to be done yet.


src/signTx.h
```diff
@@ -16,10 +17,11 @@ typedef enum {
 	SIGN_STAGE_FEE = 26,
 	SIGN_STAGE_TTL = 27,
 	SIGN_STAGE_CERTIFICATES = 28,
-	SIGN_STAGE_WITHDRAWALS = 29,
-	SIGN_STAGE_METADATA = 30,
-	SIGN_STAGE_CONFIRM = 31,
-	SIGN_STAGE_WITNESSES = 32,
+	SIGN_STAGE_CERTIFICATES_PRC = 29, // pool registration certificate sub-machine
```
refi93, line 20: what this "PRC" suffix stands for?

src/uiScreens.c
```diff
@@ -215,3 +220,152 @@ void ui_displayNetworkParamsScreen(
 	        callback
 	);
 }
+
+void ui_displayHexBufferScreen(
```
refi93, line 224: I'm wondering, why can't we just reuse the logic to display the pool id from stakepool delegation ui (signTx_handleCertificate_ui_runStep)?

src/signTxPoolRegistration.c
```diff
@@ -0,0 +1,836 @@
...
+// ============================== METADATA ==============================
+
+enum {
+	HANDLE_METADATA_STEP_DISPLAY_URL = 6340,
+	HANDLE_METADATA_STEP_DISPLAY_HASH,
+	HANDLE_METADATA_STEP_RESPOND,
+	HANDLE_METADATA_STEP_INVALID,
+};
+
+static void handleMetadata_ui_runStep()
```
refi93, line 585: It probably makes sense to let the user know if they are signing a private pool (i.e. with null metadata), at least that's what we are doing on Trezor

src/signTxPoolRegistration.c
```diff
@@ -0,0 +1,836 @@
...
+		ASSERT_TYPE(subctx->numOwners, uint16_t);
+		ASSERT_TYPE(subctx->numRelays, uint16_t);
+		subctx->numOwners = (uint16_t) u4be_read(wireHeader->numOwners);
+		subctx->numRelays = (uint16_t) u4be_read(wireHeader->numRelays);
+
+		TRACE(
+		        "num owners, relays: %d %d",
+		        subctx->numOwners, subctx->numRelays
+		);
+		VALIDATE(subctx->numOwners < POOL_MAX_OWNERS, ERR_INVALID_DATA);
```
refi93, line 269: shouldn't this technically be a `<=` comparison? I mean if something is declared to be the "MAX" value, I would expect the max itself to be a valid value

src/signTx.c
```diff
@@ -250,6 +313,10 @@ static void signTx_handleInitAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
 		VALIDATE(ctx->numCertificates < SIGN_MAX_CERTIFICATES, ERR_INVALID_DATA);
 		VALIDATE(ctx->numWithdrawals < SIGN_MAX_REWARD_WITHDRAWALS, ERR_INVALID_DATA);
 
+		// TODO isn't this more like a security policy?
+		if (ctx->isSigningPoolRegistrationAsOwner)
+			VALIDATE(ctx->numCertificates == 1, ERR_INVALID_DATA);
```
refi93, line 318: I would expect here an enforcement of numWithdrawals being zero as well. How is this ensured right now?

src/signTxPoolRegistration.c
```diff
@@ -0,0 +1,836 @@
...
+			view_skipBytes(&view, bip44_parseFromWire(&subctx->owner.path, VIEW_REMAINING_TO_TUPLE_BUF_SIZE(&view)));
+			// further validation of the path in security policy below
+			TRACE("Owner given by path:");
+			BIP44_PRINTF(&subctx->owner.path);
+			VALIDATE(view_remainingSize(&view) == 0, ERR_INVALID_DATA);
+
+			subctx->numOwnersGivenByPath++;
+			VALIDATE(subctx->numOwnersGivenByPath <= 1, ERR_INVALID_DATA);
+
+			break;
```
refi93, line 370: I know for an unsuported "owner type" the UI logic would fail, but for the sake of clarity/"foolproofness", I would expect this switch to throw if an unsupported pool owner is passed

src/signTxPoolRegistration.c
```diff
@@ -0,0 +1,836 @@
...
+			VALIDATE(view_remainingSize(&view) <= 64, ERR_INVALID_DATA);
+			str_validateText(VIEW_REMAINING_TO_TUPLE_BUF_SIZE(&view));
+
+			TRACE("Adding relay format 2 to tx hash");
+			txHashBuilder_addPoolRegistrationCertificate_addRelay2(
+			        txHashBuilder,
+			        VIEW_REMAINING_TO_TUPLE_BUF_SIZE(&view)
+			);
+			break;
+		}
```
refi93, line 559: shouldn't this throw if an unsupported relay type is provided?

src/textUtils.c
```diff
@@ -101,7 +116,65 @@ size_t str_formatTtl(uint64_t ttl, char* out, size_t outSize)
...
+#ifdef DEVEL
+// only used in internal device tests
+size_t urlToBuffer(const char* url, uint8_t* buffer, size_t bufferSize)
+{
+	size_t urlLength = strlen(url);
+	ASSERT(urlLength < BUFFER_SIZE_PARANOIA);
+	ASSERT(bufferSize < BUFFER_SIZE_PARANOIA);
+	ASSERT(bufferSize >= urlLength);
+
+	const char* validChars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-._~:/?#[]@!$&'()*+,;%=";
```
refi93, line 145: is there any point in validating the input for a function used exclusively for tests?

src/txHashBuilder_test.c
```diff
@@ -123,13 +125,14 @@
...
...139C8F1740FFB8E7AA9E5232DC82018200581C337B62CFFF6403A06A3ACBC34F8C46003C69FE79A3628CEFA9C472518A03581C5631EDE662CFB10FD5FD69B4667101DD289568E12BCF5F64D1C406FC5820198890AD6C92E80FBDAB554DDA02DA9FB49D001BBD96181F3E07F7A6AB0D06401A1DCD65001A1443FD00D81E820101581DE03A7F09D3DF4CF66A7399C2B05BFA234D5A29560C311FC5DB4C49071181581C3A7F09D3DF4CF66A7399C2B05BFA234D5A29560C311FC5DB4C4907118384001904D24408080808F68301F678204141414120323430303A636230303A323034393A313A3A613239663A31383034820278204141414120323430303A636230303A323034393A313A3A613239663A3138303482781968747470733A2F2F746573747374616B65706F6F6C2E636F6D5820914C57C1F12BBF4A82B12D977D4F274674856A11ED4B9B95BD70F5D41C5064A683028200581C32C728D3861E164CAB28CB8F006448139C8F1740FFB8E7AA9E5232DC581C0D13015CDBCDBD0889CE276192A1601F2D4D20B8392D4EF4F9A754E283028200581C32C728D3861E164CAB28CB8F006448139C8F1740FFB8E7AA9E5232DC581C1D13015CDBCDBD0889CE276192A1601F2D4D20B8392D4EF4F9A754E283028200581C32C728D3861E164CAB28CB8F006448139C8F1740FFB8E7AA9E5232DC581C2D13015CDBCDBD0889CE276192A1601F2D4D20B8392D4EF4F9A754E205A1581DE032C728D3861E164CAB28CB8F006448139C8F1740FFB8E7AA9E5232DC19029A075820DEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF
```
refi93, line 128: for the sake of consistency, this should be lowercase

src/txHashBuilder_test.c
```diff
@@ -225,16 +228,89 @@ static struct {
...
+	size_t ipv4Size = decode_hex("08080808", ipv4, 4);
+	ASSERT(ipv4Size == 4);
+	uint16_t port = 1234;
+
+	txHashBuilder_addPoolRegistrationCertificate_addRelay0(builder, &port, ipv4, ipv4Size, NULL, 0);
+
+	const char* dnsName = "AAAA 2400:cb00:2049:1::a29f:1804";
+	uint8_t dnsNameBuffer[64];
+	size_t dnsNameBufferSize = dnsNameToBuffer(dnsName, dnsNameBuffer, SIZEOF(dnsNameBuffer));
+	ASSERT(dnsNameBufferSize <= SIZEOF(dnsNameBuffer));
```
refi93, line 286: I would expect a test assertion to match on some exact value, not just a threshold, otherwise it is unlikely to catch implementation changes

src/txHashBuilder_test.c
```diff
@@ -225,16 +228,89 @@ static struct {
...
+	txHashBuilder_addPoolRegistrationCertificate_enterRelays(builder);
+
+	uint8_t ipv4[4];
+	size_t ipv4Size = decode_hex("08080808", ipv4, 4);
+	ASSERT(ipv4Size == 4);
+	uint16_t port = 1234;
+
+	txHashBuilder_addPoolRegistrationCertificate_addRelay0(builder, &port, ipv4, ipv4Size, NULL, 0);
+
+	const char* dnsName = "AAAA 2400:cb00:2049:1::a29f:1804";
```
refi93, line 283: for the sake of tests I would pick a legit-looking dns name e.g. `test.test.sth` or something like that , otherwise this has the potential to confuse a casual reader of the code

janmazak: This actually is a legit AAAA DNS record. That `dns_name` is probably a misnomer; CDDL says 
`dns_name ; An A or AAAA DNS record`, so it does not seem to be a name. Perhaps I just add an explanatory comment?

src/signTx.h
janmazak: Pool Registration Certificate

src/uiScreens.c
janmazak: Well, I don't see an elegant way to do it. Manipulation with strings is quite clumsy in C. It would be nice if I could display pool id for stake delegation certificate in the same place in code as for other certificates, but it only arrives in the following APDU in the sub-machine (I can send it in the certificate one, but then I have to store it, avoid having two different copies in two places (I definitely want to avoid passing pool params to the tx hash builder separately), and the approach would not be very consistent --- showing some parameters in the main machine, some in the sub-machine).

src/textUtils.c
```diff
@@ -101,7 +116,65 @@ size_t str_formatTtl(uint64_t ttl, char* out, size_t outSize)
...
+
+size_t dnsNameToBuffer(const char* dnsName, uint8_t* buffer, size_t bufferSize)
+{
+	size_t dnsNameLength = strlen(dnsName);
+	ASSERT(dnsNameLength < BUFFER_SIZE_PARANOIA);
+	ASSERT(bufferSize < BUFFER_SIZE_PARANOIA);
+	ASSERT(bufferSize >= dnsNameLength);
+
+	for (size_t i = 0; i < dnsNameLength; i++) {
+		if (dnsName[i] > 127 || dnsName[i] == 0)
```
ppershing, line 172: I would restrict this to more reasonable range. Check ascii table but I dont think anything below 32 should be allowed

src/textUtils.c
ppershing: That's a good question. How do we do validation in real code? Also, looking below, maybe we should have one
`validateStringContainsOnlyAllowedChars(string, char_string)` function

src/signTx.h
ppershing: I would just rename it to SIGN_STAGE_CERTIFICATES_POOL

src/txHashBuilder.c
```diff
@@ -288,20 +292,358 @@ void txHashBuilder_addCertificate_delegation(
...
+	}
+}
+
+void txHashBuilder_addPoolRegistrationCertificate_enterRelays(tx_hash_builder_t* builder)
+{
+	TRACE("txHashBuilder_addPoolRegistrationCertificate_enterRelays: %d", builder->state);
+
+	// enter empty owners if none were received
+	if (builder->state == TX_HASH_BUILDER_IN_CERTIFICATES_POOL)
+		txHashBuilder_addPoolRegistrationCertificate_enterOwners(builder);
```
ppershing, line 401: this should assert that there were no owners to start with ...

src/txHashBuilder.c
```diff
@@ -288,20 +292,358 @@ void txHashBuilder_addCertificate_delegation(
...
+        tx_hash_builder_t* builder,
+        const uint16_t* port,
+        const uint8_t* ipv4, size_t ipv4Size,
+        const uint8_t* ipv6, size_t ipv6Size
+)
+{
+	TRACE("txHashBuilder_addPoolRegistrationCertificate_addRelay0: %d", builder->state);
+
+	addRelay_updateState(builder);
+	ASSERT(builder->state == TX_HASH_BUILDER_IN_CERTIFICATES_POOL_RELAYS);
```
ppershing, line 440: Can relay0, relay1, ... be entered in *any* order?

src/txHashBuilder.c
```diff
@@ -288,20 +292,358 @@ void txHashBuilder_addCertificate_delegation(
...
+	ASSERT(builder->poolCertificateData.remainingRelays > 0);
+	builder->poolCertificateData.remainingRelays--;
+
+	ASSERT(ipv4Size < BUFFER_SIZE_PARANOIA);
```
ppershing, line 444: shouldn't we have `struct ipv4_t {ip: uint8_t[4]}` for this? Similarly for ipv6

src/txHashBuilder.c
```diff
@@ -288,20 +292,358 @@ void txHashBuilder_addCertificate_delegation(
...
+			}
+		}
+		{
+			if (ipv6 != NULL) {
+				ASSERT(ipv6Size == 16);
+				BUILDER_APPEND_CBOR(CBOR_TYPE_BYTES, ipv6Size);
+
+				// serialized as 4 big-endian uint32
+				for (size_t i = 0; i < 4; i++) {
+					for (size_t j = 0; j < 4; j++) {
```
ppershing, line 481: You should preferably use endian-conversion utilities we have, not this hand-written code

src/txHashBuilder.c
```diff
@@ -288,20 +292,358 @@ void txHashBuilder_addCertificate_delegation(
...
+			BUILDER_APPEND_DATA(dnsName, dnsNameSize);
+		}
+	}
+}
+
+// enter empty owners or relays if none were received
+static void addPoolMetadata_updateState(tx_hash_builder_t* builder)
+{
+	switch (builder->state) {
+	case TX_HASH_BUILDER_IN_CERTIFICATES_POOL:
```
ppershing, line 572: thouldn't there be enter pool owners?

src/uiScreens.c
```diff
@@ -215,3 +220,152 @@ void ui_displayNetworkParamsScreen(
...
+
+	if (owner->ownerType == SIGN_TX_POOL_OWNER_TYPE_PATH) {
+		bip44_printToStr(&owner->path, ownerDescription, SIZEOF(ownerDescription));
+	}
+
+	{
+		// add owner (represented as bech32-encoded reward account for owner's staking key)
+		size_t pathLength = strlen(ownerDescription);
+		ASSERT(pathLength <= MAX_PATH_LENGTH);
+		ASSERT(pathLength <= SIZEOF(ownerDescription) - 2);
```
ppershing, line 349: prefer + instead of - as this forces unsigned->signed conversion chaos

src/uiScreens.c
```diff
@@ -215,3 +220,152 @@ void ui_displayNetworkParamsScreen(
...
+
+		if (pathLength > 0) {
+			// add a space
+			ownerDescription[pathLength++] = ' ';
+			ASSERT(pathLength <= SIZEOF(ownerDescription) - 1);
+			ownerDescription[pathLength] = '\0';
+			ASSERT(pathLength == strlen(ownerDescription));
+		}
+
+		size_t ownerAccountLength = humanReadableAddress(
```
ppershing, line 359: I think I would prefer style `strLen += humanRe...()`, that way you can perform multiple appends with

```
len += strAppend(str+len, ..., )
len += strAppend(...., )
```

src/txHashBuilder.c
janmazak: I am not aware of any restriction on relay ordering. We are adding them to the tx body as they come (each in its own APDU). We are not doing much with the relays (not displayed to the user and not fully validated); the sw wallet is responsible for giving us something valid.

src/txHashBuilder_test.c
refi93: interesting, let's keep this then and leave o comment clarifying it's a sample AAAA DNS record

src/signTxPoolRegistration.c
janmazak: Yes, though this is the way how it was done previously for inputs etc. I have changed it in other places too.

src/txHashBuilder_test.c
janmazak: I'd rather keep it as it is, since uppercase is what comes out of the ledger via the usb tool (perhaps that could be changed, but likely not worth the trouble) and what is used for comparison. We already have a mix of upper- and lowercase in the file, depending on where the data come from.

src/txHashBuilder.c
janmazak: Fixed in f8a3823e9908541efd76425fa691875a4867ed1a.

src/txHashBuilder.c
janmazak: Something like
```
					uint8_t chunk[4];
					uint32_t* valuePtr = (uint32_t *)(ipv6 + 4 * i);
					u4be_write(chunk, *valuePtr);
```
?

src/txHashBuilder_test.c
janmazak: I've found that the upper/lowercase was not consistent (mostly lowercase), so I ended up changing this.

src/textUtils.c
janmazak: Done in 25aa9200a2c2b52fea48c4f9b51aa497be332a97.

src/textUtils.c
janmazak: This is kind of a leftover from earlier times. I rewrote it in 25aa9200a2c2b52fea48c4f9b51aa497be332a97.

refi93:
displaying transaction fee/ttl in the UI is IMHO not an essential information for an owner and we can jump directly into showing the stakepool registration data, or was it requested by IOHK?

src/textUtils.h
```diff
@@ -5,10 +5,29 @@
 
 size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize);
 
+#ifdef DEVEL
+void str_traceAdaAmount(const char *prefix, uint64_t amount);
+#define TRACE_ADA_AMOUNT(PREFIX, AMOUNT) \
```
refi93, line 10: just curious, what's the benefit of this debugging macro vs just dumping the prefix and ada amount separately with existing macros? And if it is just saving a line or two of other debugging macros, is it worth it?

janmazak: The problem is a very complicated formatting of ada amounts, a function must be called for that. And you have to allocate and pass a buffer to the function, and combining it with the prefix is annoying in C.

janmazak:
All of the agreed-upon UI changes are now implemented.

src/textUtils.c
```diff
@@ -101,7 +116,40 @@ size_t str_formatTtl(uint64_t ttl, char* out, size_t outSize)
...
+	for (size_t i = 0; i < textSize; i++) {
+		VALIDATE(text[i] <= 126, ERR_INVALID_DATA);
+		VALIDATE(text[i] >= 32, ERR_INVALID_DATA);
+	}
+}
+
+#ifdef DEVEL
+
+// converts a text to bytes (suitable for CBORization) and validates if chars are allowed
+size_t str_textToBuffer(const char* text, uint8_t* buffer, size_t bufferSize)
```
refi93, line 139: I would suggest separating the validation and "cborization" logic and do the validation explicitly wherever it makes sense. But I see that this helper function is used only for tests, so I would drop the char code validation from there completely - there's imho not a point in validating test fixtures if they would not be validated in the actual logic either.

And if anybody deems this function reusable in the future for actual logic, they should still do due diligence and implement/use whatever input validation makes sense for their use case (possibly this one, possibly other)

refi93 (changes requested):

src/signTx.h
```diff
@@ -62,8 +69,14 @@ typedef struct {
 typedef struct {
 	sign_tx_stage_t stage;
 
-	uint8_t networkId; // part of Shelley address
-	uint32_t protocolMagic; // part of Byron address
+	// the presence of a stake pool registration certificate
+	// significantly affects restrictions on the whole tx
+	bool isSigningPoolRegistrationAsOwner;
+
+	// TODO FIXME add variables for receivedFee, receivedTtl, (receivedMetadata?) and move fee and ttl to stagedata ?
```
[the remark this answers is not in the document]
janmazak: Yes, it's done now, see e1376ed3f7e99da6435b046268ff88b445bdc751.

src/textUtils.c
janmazak: Fixtures are now validated in precisely the same way as actual data, so I'm keeping the validation here. 

(Generally, I have messed up test fixtures many times, so I do feel some need to validate them whenever it is easy and cheap.)

refi93 (approved):

src/uiScreens.c
```diff
@@ -215,3 +218,161 @@ void ui_displayNetworkParamsScreen(
...
+				.type = REWARD,
+				.networkId = networkId,
+				.spendingKeyPath = owner->path,
+				.stakingChoice = NO_STAKING,
+			};
+
+			deriveAddress(
+			        &rewardAddressParams,
+			        rewardAddress,
+			        SIZEOF(rewardAddress)
```
ppershing, line 328: can we have this on the same line as rewardAddress ?

src/txHashBuilder.c
```diff
@@ -288,20 +292,347 @@ void txHashBuilder_addCertificate_delegation(
...
+		}
+		{
+			if (ipv6 != NULL) {
+				ASSERT(sizeof(ipv6->ip) == IPV6_SIZE); // SIZEOF does not work for 4-byte buffers
+				BUILDER_APPEND_CBOR(CBOR_TYPE_BYTES, IPV6_SIZE);
+
+				// serialized as 4 big-endian uint32
+				for (size_t i = 0; i < 4; i++) {
+					uint8_t chunk[4];
+					uint32_t* valuePtr = (uint32_t *)(ipv6->ip + 4 * i);
```
ppershing, line 463: could be `uint32_t* as_uint32 = ipv6->ip` and then index directly into `as_uint32`

src/txHashBuilder.c
```diff
@@ -288,20 +292,347 @@ void txHashBuilder_addCertificate_delegation(
...
+	ASSERT(builder->state == TX_HASH_BUILDER_IN_CERTIFICATES_POOL_METADATA);
+	{
+		BUILDER_APPEND_CBOR(CBOR_TYPE_NULL, 0);
+	}
+	builder->state = TX_HASH_BUILDER_IN_CERTIFICATES;
+}
+
+static void txHashBuilder_assertCanLeaveCertificates(tx_hash_builder_t* builder)
+{
 	if (builder->state == TX_HASH_BUILDER_IN_CERTIFICATES) {
```
ppershing, line 622: I think I like the `switch` variation you introduced in addPoolMetadata_updateState nicer to read

src/signTx.c
```diff
@@ -120,11 +122,59 @@ static inline void advanceStage()
...
+// called from main state machine when a pool registration certificate
+// sub-machine is finished, or when other type of certificate is processed
+static inline void advanceCertificatesStateIfAppropriate()
 {
-	TRACE();
-	io_send_buf(SUCCESS, NULL, 0);
-	ui_displayBusy(); // needs to happen after I/O
+	TRACE("%u", ctx->stage);
+
+	if (ctx->stage == SIGN_STAGE_CERTIFICATES) {
```
ppershing, line 131: here I would likewise prefer switch on ctx->stage instead of assert in else branch

src/signTxPoolRegistration.c
```diff
@@ -0,0 +1,878 @@
...
+	UI_STEP_BEGIN(subctx->ui_step);
+
+	UI_STEP(HANDLE_OWNER_STEP_DISPLAY) {
+		ui_displayOwnerScreen(&subctx->owner, subctx->currentOwner, commonTxData->networkId, this_fn);
+	}
+	UI_STEP(HANDLE_OWNER_STEP_RESPOND) {
+		respondSuccessEmptyMsg();
+
+		subctx->currentOwner++;
+		if (subctx->currentOwner == subctx->numOwners) {
```
ppershing, line 315: isn't this included in advanceState? :thinking:

ppershing (approved):
Quite good job!

src/signTxPoolRegistration.c
janmazak: My preference is to avoid validation in advanceState, so there are just asserts making sure that an invalid situation has been already properly detected when handling APDUs (in this case, a missing/superfluous path owner).

PR #42, opened 2020-11-05
WIP: bulk key export

janmazak:
Idea: a single UI state machine and key derivation routine (`returnOneKey`), with two different entry points (bulk and single-key) and slightly different security policies (if bulk export, i.e. pre-confirmed, no more confirmations are asked --- not implemented yet).

Works with legacy APDU flow, we make a distinction on the parameter p1 in APDU (consistent with what we do in signTx --- p1 is used to describe the stage).


src/getPublicKeys.c
```diff
@@ -0,0 +1,284 @@
...
+			return false;
+
+		case GET_KEYS_STAGE_GET_KEYS:
+			return ctx->remainingKeys > 0;
+		
+		default:
+			ASSERT(false);
+	}
+	ASSERT(false);
+	return false;
```
refi93, line 38: if the switch covers all cases and even asserts on default case, this is imho more confusing than beneficial as assurance, or was this pattern already adopted in other parts of the codebase?

src/getPublicKeys.c
```diff
@@ -0,0 +1,284 @@
...
+
+		CASE(0x01, getPublicKeys_handleInitAPDU);
+		CASE(0x02, getPublicKeys_handleGetKeyAPDU);
+		DEFAULT(NULL)
+#	undef   CASE
+#	undef   DEFAULT
+	}
+}
+
+void getPublicKeys_handleAPDU(
```
refi93, line 267: is this function used anywhere?

janmazak: Yes, it's the main entry point used in `handlers.c`.

src/getPublicKeys.c
```diff
@@ -0,0 +1,265 @@
...
+)
+{
+    CHECK_STAGE(GET_KEYS_STAGE_GET_KEYS);
+
+    ASSERT(wireDataSize < BUFFER_SIZE_PARANOIA);
+
+    VALIDATE(ctx->remainingKeys >= 1, ERR_INVALID_STATE);
+    ctx->remainingKeys--;
+
+	read_view_t view = make_read_view(wireDataBuffer, wireDataBuffer + wireDataSize);
```
ppershing, line 224: this is some weird indent. is the code formatted properly?

src/getPublicKeys.c
```diff
@@ -0,0 +1,265 @@
...
+	{
+		// parse data
+
+		TRACE_BUFFER(wireDataBuffer, wireDataSize);
+		read_view_t view = make_read_view(wireDataBuffer, wireDataBuffer + wireDataSize);
+
+		parsePath(&view);
+
+		size_t remaining = view_remainingSize(&view);
+		switch (remaining) {
```
ppershing, line 159: what's the point of making this backwards compatible? We own both JS and device and so JS can be responsible for proper interaction

ppershing:
I am quite confused. Do we have somewhere documentation what is the expected user interaction with the device given single and bulk key export and given different paths you want to export?

PR #43, opened 2020-11-08
add support for bulk key export

janmazak:
If I have not missed something, it should be complete.


src/getPublicKeys.h
```diff
@@ -0,0 +1,33 @@
...
+#include "bip44.h"
+#include "keyDerivation.h"
+
+#define MAX_PUBLIC_KEYS 1000
+
+typedef enum {
+	GET_KEYS_STAGE_NONE = 0,
+	GET_KEYS_STAGE_INIT = 20,
+	GET_KEYS_STAGE_GET_KEYS = 40,
+	GET_KEYS_STAGE_SINGLE_KEY = 60 // terminal state for the case where a single key is asked
```
refi93, line 15: I don't see this stage being referenced anywhere else, is it just a leftover from the previous version?

src/getPublicKeys.h
```diff
@@ -0,0 +1,33 @@
...
+} get_keys_stage_t;
+
+typedef struct {
+	get_keys_stage_t stage;
+
+	uint16_t remainingKeys;
```
refi93, line 21: I think `remainingPaths` would be more accurate as the "key" is just a product of the call, technically

refi93 (approved):

src/getPublicKeys.c
```diff
@@ -0,0 +1,265 @@
...
+		} else {
+			ui_idle(); // we are done, display the main app screen
+		}
+	}
+	UI_STEP_END(GET_KEY_UI_STEP_INVALID);
+}
+
+// derive the key described by ctx->pathSpec and run the ui state machine accordingly
+// if isSingleKeyExport is true, more restrictive security policy is applied
+static void returnOneKey(bool isSingleKeyExport)
```
ppershing, line 75: can we replace boolean with enum?

ppershing: Actually, why don't we
1) rename this to displayPromptAndRespond() or something like that?
2) it would make sense to just pass security_policy to this function instead of passing in boolean/enum

src/getPublicKeys.c
```diff
@@ -0,0 +1,265 @@
...
+		        secondLine,
+		        this_fn,
+		        respond_with_user_reject
+		);
+	}
+	UI_STEP(HANDLE_INIT_UI_STEP_RESPOND) {
+		ASSERT(ctx->remainingPaths > 0);
+		returnOneKey(false); // runs another UI state machine
+
+		ctx->stage = GET_KEYS_STAGE_GET_KEYS;
```
ppershing, line 138: I am worried about this. returnOneKey is async and I don't know what are the consequences of setting this afterwards ...

src/getPublicKeys.c
```diff
@@ -0,0 +1,265 @@
...
+			ctx->remainingPaths = (uint16_t) remainingPaths;
+			break;
+		}
+		default: {
+			THROW(ERR_INVALID_DATA);
+		}
+		}
+	}
+
+	if (ctx->remainingPaths == 0) {
```
ppershing, line 183: why do we have this fast-track? Especially because it is skipping the policy below...

src/getPublicKeys.c
```diff
@@ -0,0 +1,265 @@
...
+		ui_displayPrompt(
+		        "Confirm export",
+		        secondLine,
+		        this_fn,
+		        respond_with_user_reject
+		);
+	}
+	UI_STEP(HANDLE_INIT_UI_STEP_RESPOND) {
+		ASSERT(ctx->remainingPaths > 0);
+		returnOneKey(false); // runs another UI state machine
```
ppershing, line 136: I am little worried here. returnOneKey *messes up* `ctx->ui_step`. While behavior is technically correct, it is really dangeours if somebody does a negligent update of the code in the future. We should somehow warn about this (maybe also by improving naming of the function)

src/getPublicKeys.c
refi93: it skips the prompt for the count of exported keys, but given that `policyForGetPublicKeysInit` takes as a parameter the remaining paths count, it probably makes more sense to move the check for zero there and return POLICY_ALLOW_WITHOUT_PROMPT in that case

refi93: ahh, `returnOneKey` returns the key already (something that would be done for the second time with `getPublicKeys_handleInit_ui_runStep`

refi93: but it seems to me that we should be able to get rid of this fast-track, if for `remainingPaths == 0` we had `POLICY_ALLOW_WITHOUT_PROMPT` policy here, we would keep in ctx apart from "remainingPaths" also the "totalPaths" count which would be the count of total keys to be returned from this flow (i.e. won't decrease throughout the flow) and then the `returnOneKey's` policy would be `POLICY_PROMPT_BEFORE_RESPONSE` if `totalPaths` equals zero, and there would be no need to pass any bool or policy to returnOne key as all the info needed to evaluate the policy (i.e. whether the flow is a single or bulk key export, based on the totalPaths value) would live within the context

src/getPublicKeys.c
```diff
@@ -0,0 +1,265 @@
...
+static void getPublicKeys_handleInit_ui_runStep()
+{
+	TRACE("UI step %d", ctx->ui_step);
+	ui_callback_fn_t* this_fn = getPublicKeys_handleInit_ui_runStep;
+
+	UI_STEP_BEGIN(ctx->ui_step);
+	UI_STEP(HANDLE_INIT_UI_STEP_CONFIRM) {
+		char secondLine[100];
+		explicit_bzero(secondLine, SIZEOF(secondLine));
+		snprintf(secondLine, SIZEOF(secondLine), "%u public keys?", ctx->remainingPaths);
```
refi93, line 124: this should be `remainingPaths + 1` indeed

refi93:
If the first path supplied to the bulk export command happens to trigger a prompt, it causes Ledger to crash. I was able to cause this behavior with the following paths supplied to the bulk key export command:

```
    const paths = [
      "44'/1815'/1'/0/12'",
      "44'/1815'/1'/0/10'/1/2/3",
      "1852'/1815'/0'/0/1",
      "1852'/1815'/0'/2/0"
    ];
```

Notice that the first path is triggers the confirmation flow within `returnOneKey` function as the key is not not silently exportable by the policies

It seems to me that nested UI state machines clash on some shared variable, probably the callback pointer :/

src/getPublicKeys.c
```diff
@@ -21,6 +21,55 @@ static void parsePath(read_view_t* view)
...
+		ui_idle(); // we are done with this key export
+		break;
+
+	case SIGN_STAGE_NONE:
+	default:
+		ASSERT(false);
+	}
+}
+
+static void advanceStageIfAppropriate()
```
refi93, line 53: I would prefer to put the condition `if (ctx->currentPath == ctx->numPaths) ` right in the part of code where this function is invoked. I see similar conditional calls of advanceStage in the signTx.c file so it should be consistent with the existing codebase as well and IMHO more readable unless I'm missing something

src/getPublicKeys.c
```diff
@@ -228,7 +275,7 @@ void getPublicKeys_handleGetNextKeyAPDU(
 	parsePath(&view);
 	VALIDATE(view_remainingSize(&view) == 0, ERR_INVALID_DATA);
 
-	returnOneKey();
+	promptAndRespondOneKey();
```
refi93, line 278: the "prompt" verb may be imho misleading given that the prompt is triggered only under certain condiditons. Imho a name like  `getPublicKeyFlow()` would be more appropriate as it gives the information to the reader that it will get one public key and some user-interaction may or may not be happening as it is a flow

src/getPublicKeys.c
janmazak: The difference (vs. signTx) is that one has to distinguish if we are in _INIT or _GET_KEY state. Essentially, the body of this function can be copied to the place where it is called (there is just one), but I like to keep the code inside UI state machines as short as possible (like, a single screen), otherwise it becomes hard to read thanks to the unbalanced macros.

refi93 (approved):

refi93 (approved):

src/getPublicKeys.c
ppershing: why do we actually have two functions? advanceStage seems a bit redundant to me :thinking:

janmazak: This was already refactored, there is just one function (advanceStage) now.

PR #45, opened 2020-11-24
allow unreasonable accounts for staking paths

janmazak:


src/securityPolicy.c
```diff
@@ -364,6 +367,8 @@ security_policy_t policyForSignTxWitness(
 
 	if (isSigningPoolRegistrationAsOwner) {
 		DENY_UNLESS(is_valid_stake_pool_owner_path(pathSpec));
+
+		// TODO warn if account is not reasonable? see also below, we should once and for all fix the TODOs there
```
refi93, line 371: imho it makes sense to warn here and I would just drop the TODOs below - imho there's no harm in warning the users, the policies are clear on their own and I don't see an added value in the comments, it may even be counterproductive for anybody that would like to revisit the policies in the future as it may move the focus away from arguably a lot more relevant ones

refi93: indeed, `WARN_IF(is_too_deep(pathSpec))` should imho be DENY_IF - all other occurences of that check result in a denial, so I would align that. It's kind of weird to forbid the path to be used as change but allow spending from it

refi93 (changes requested):
see comments where I left suggestions on how I would deal with the TODOs and finish the logic

src/securityPolicy.c
```diff
@@ -362,14 +365,18 @@ security_policy_t policyForSignTxWitness(
 {
 	DENY_UNLESS(is_valid_witness(pathSpec));
 
+	WARN_UNLESS(has_reasonable_account(pathSpec));
```
refi93, line 368: won't this technically make it possible (with a warning though) to sign a stake pool registration with a spending key for an "unreasonable" account number?

refi93: how I imagine the logic:

```
DENY_UNLESS(is_valid_witness(pathSpec));

if (isSigningPoolRegistrationAsOwner) {
  DENY_UNLESS(is_valid_stake_pool_owner_path(pathSpec));
}
DENY_IF(is_too_deep(pathSpec));
WARN_UNLESS(has_reasonable_account_and_address(pathSpec));
```

i.e. I think the deny policies should come first, otherwise they can be invalidated by the WARN_IF policies if they happen to overlap (and the deny action should probably have precedence)

src/securityPolicy.c
```diff
@@ -363,13 +366,17 @@ security_policy_t policyForSignTxWitness(
...
+
 	} else {
-		// TODO Perhaps we can relax this?
+		// witness for input or withdrawal
+
 		WARN_UNLESS(has_reasonable_account_and_address(pathSpec));
 
-		// TODO deny this? or check for depth in is_valid_witness?
-		WARN_IF(is_too_deep(pathSpec));
+		DENY_IF(is_too_deep(pathSpec));
```
refi93, line 379: If this comes after `WARN_UNLESS`, wouldn't that mean that it deep paths with "unreasonable" account and address index would pass with a warning instead of being denied?

src/securityPolicy.c
```diff
@@ -363,13 +366,17 @@ security_policy_t policyForSignTxWitness(
 	DENY_UNLESS(is_valid_witness(pathSpec));
 
 	if (isSigningPoolRegistrationAsOwner) {
+
 		DENY_UNLESS(is_valid_stake_pool_owner_path(pathSpec));
+
+		WARN_UNLESS(bip44_hasReasonableAccount(pathSpec));
```
refi93, line 372: I would just do `WARN_UNLESS(has_reasonable_account_and_address(pathSpec));` for both branches as we apply this logic for staking paths if isSigningPoolRegistrationAsOwner is false and I see no reason to do something different for this usecase, even though the address index check is technically redundant, but I would opt for consistency and simplicity here

src/securityPolicy.c
```diff
@@ -365,13 +368,12 @@ security_policy_t policyForSignTxWitness(
 	if (isSigningPoolRegistrationAsOwner) {
 		DENY_UNLESS(is_valid_stake_pool_owner_path(pathSpec));
 	} else {
-		// TODO Perhaps we can relax this?
-		WARN_UNLESS(has_reasonable_account_and_address(pathSpec));
-
-		// TODO deny this? or check for depth in is_valid_witness?
-		WARN_IF(is_too_deep(pathSpec));
+		// witness for input or withdrawal
```
refi93, line 371: also stake delegation/key deregistration is included in this branch which do require respective staking key witnesses, I would just drop the comment as it's rather self-evident.

I would also consider removing the else branch and apply the `DENY_IF(is_too_deep(pathSpec))` regardless of isSigningPoolRegistrationAsOwner value. The check is already done in the is_valid_stake_pool_owner_path so it's technically redundant but also the final code is simpler imho. Sorry, I should have realized in the previous refactoring suggestion I had

refi93 (approved):
two small suggestions and otherwise LGTM

PR #47, opened 2020-12-01
allow no outputs in signTx

janmazak:
No changes in js code are needed --- the validation did not enforce an output to be present. 
(I added a test for this case, see https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/16.)

WARNING: a tx without outputs does not include network id / protocol magic, but we do display the values at the initiation of tx signing to the user. It has not been decided yet what exactly to do about it, or whether we should add an additional warning for the case no outputs are present. Until these UI issues are resolved and implemented, this is not finished.

### DIFF
--- a/src/signTx.c
+++ b/src/signTx.c
@@ -48,7 +48,13 @@ static inline void advanceStage()
 		ASSERT(ctx->currentInput == ctx->numInputs);
 		txHashBuilder_enterOutputs(&ctx->txHashBuilder);
 		ctx->stage = SIGN_STAGE_OUTPUTS;
-		break;
+
+		if (ctx->numOutputs > 0) {
+			// wait for output APDUs
+			break;
+		}
+
+	// intentional fallthrough
 
 	case SIGN_STAGE_OUTPUTS:
 		// we should have received all outputs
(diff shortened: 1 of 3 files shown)


janmazak:
I've added a suggested change to the UI (no final decision on the matter yet).

src/signTx.c
```diff
@@ -48,7 +48,13 @@ static inline void advanceStage()
 		ASSERT(ctx->currentInput == ctx->numInputs);
 		txHashBuilder_enterOutputs(&ctx->txHashBuilder);
 		ctx->stage = SIGN_STAGE_OUTPUTS;
-		break;
+
+		if (ctx->numOutputs > 0) {
```
refi93, line 52: the CDDL specifies that the outputs key is obligatory: https://github.com/input-output-hk/cardano-ledger-specs/blob/6e30ecf73c9296bc320d71f3a6b7bfad3ce117e9/shelley/chain-and-ledger/shelley-spec-ledger-test/cddl-files/shelley.cddl#L53 - I guess this code results in omitting the key completely, right?

janmazak: We are always  calling `txHashBuilder_enterOutputs` (see a couple of lines above) which is responsible for this.

src/signTx.c
```diff
@@ -198,11 +204,29 @@ static void signTx_handleInit_ui_runStep()
 
 	UI_STEP_BEGIN(ctx->ui_step);
 	UI_STEP(HANDLE_INIT_STEP_DISPLAY_DETAILS) {
-		ui_displayNetworkParamsScreen(
-		        "New transaction",
-		        ctx->commonTxData.networkId, ctx->commonTxData.protocolMagic,
-		        this_fn
-		);
+		if (ctx->numOutputs > 0) {
```
refi93, line 207: HANDLE_INIT_STEP_DISPLAY_DETAILS is triggered when there is some kind of mismatch between protocol magic an network id. A simple fix would be adding the same condition being checked here into the policy. A caveat though is that If there is actually a mismatch between network id and protocol magic, this UI logic would skip the displaying of network id/protocol magic if the tx has a withdrawal and no outputs (where network id is actually verifiable)

https://github.com/vacuumlabs/ledger-app-cardano-shelley/blob/5f602944dd268e6363474c21858a4136f5501f39/src/securityPolicy.c#L179

refi93: pushed commit with fix

PR #48, opened 2020-12-09
Operational certificate + pool cold key derivation

refi93:
Motivation: Add instructions for derivation of pool cold keys and operational certificate signing.

Changes:
* introduced `getExtendedPoolColdPublicKey(coldKeyPath)` call - the expected derivation path structure is `m/1853'/1815'/0/<cold key index>'`
* introduced `signOperationalCertificate(kesPublicKeyHex, kesPeriod, counter, coldKeyPath)` which assembles the operational certificate and returns its signature - as the OpCert structure has no variable length fields and is well under 255 bytes, it's done in a single APDU
* the security policy both for key derivation and operational certificate signing validates the coldKey derivation path
* refactored messageSigning.c to have a common function for signing a message with a key derived from a path and purpose-specific functions for tx witnesses and operational certificate signatures with corresponding message structure and derivation path validation, to ensure pool cold keys are not used for tx witnesses and vice-versa to prevent even hypothetical collisions between the different types of messages (txs vs operational certificates) being signed by those keys
* refactored `bip44_hasValidCardanoPrefix()` into `bip44_hasValidCardanoWalletPrefix()` to introduce equivalent function for pool cold keys named `bip44_hasValidCardanoPoolColdKeyPrefix()` - which is technically not required to have but I see it as good for readability and consistency, making it clear to a casual reader of the codebase that pool cold keys and wallet keys are logically separate
* added `str_formatUint64()` helper function which prints uint64 integers in decimal representation
* added TRACE_UINT64 debugging macro to trace uint64 values like issueCounter or kesPeriod

Testing:
* ran integration tests from https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/15
```
yarn test-integration --grep signOperationalCertificate
yarn test-integration --grep getExtendedPoolColdPublicKey
```

TODO: add docs (to be added once high-level idea is validated)
TODO2: missing flag in makefile to separate operator functionality (IMHO worth doing after we merge the with the rest of functionality)


src/signOpCert.c
```diff
@@ -0,0 +1,169 @@
...
+		explicit_bzero(ctx, SIZEOF(*ctx));
+	}
+	ctx->responseReadyMagic = 0;
+
+	// Validate params
+	VALIDATE(p1 == P1_UNUSED, ERR_INVALID_REQUEST_PARAMETERS);
+	VALIDATE(p2 == P2_UNUSED, ERR_INVALID_REQUEST_PARAMETERS);
+
+	{
+		TRACE_BUFFER(wireDataBuffer, wireDataSize);
```
janmazak, line 50: For fixed-sized stuff we usually use the "wireheader" version: I think the reason to use it is because it nicely lists all the required things in a single place.
see e.g. signTx_handleInitAPDU

src/signOpCert.c
```diff
@@ -0,0 +1,169 @@
...
+		read_view_t view = make_read_view(wireDataBuffer, wireDataBuffer + wireDataSize);
+
+		STATIC_ASSERT(SIZEOF(ctx->kesPublicKey) == KES_PUBLIC_KEY_LENGTH, "wrong KES public key size");
+		VALIDATE(view_remainingSize(&view) >= KES_PUBLIC_KEY_LENGTH, ERR_INVALID_DATA);
+		view_memmove(ctx->kesPublicKey, view.ptr, KES_PUBLIC_KEY_LENGTH);
```
janmazak, line 55: I find it often useful if individual parsed values are traced, too.

src/signOpCert.c
```diff
@@ -0,0 +1,169 @@
...
+	ctx->responseReadyMagic = RESPONSE_READY_MAGIC;
+
+	switch (policy) {
+#	define  CASE(policy, step) case policy: {ctx->ui_step = step; break;}
+		CASE(POLICY_PROMPT_WARN_UNUSUAL,    UI_STEP_WARNING);
+		CASE(POLICY_PROMPT_BEFORE_RESPONSE, UI_STEP_CONFIRM_START);
+		CASE(POLICY_ALLOW_WITHOUT_PROMPT,   UI_STEP_RESPOND);
+#	undef   CASE
+	default:
+		ASSERT(false);
```
janmazak, line 99: In other places, we use 
			THROW(ERR_NOT_IMPLEMENTED);
I'm not sure which is better or if we should do the same everywhere.

src/signOpCert.c
```diff
@@ -0,0 +1,169 @@
...
+	UI_STEP(UI_STEP_DISPLAY_KES_PUBLIC_KEY) {
+		ui_displayHexBufferScreen(
+		        "KES public key",
+		        ctx->kesPublicKey, SIZEOF(ctx->kesPublicKey),
+		        this_fn
+		);
+	}
+	UI_STEP(UI_STEP_DISPLAY_KES_PERIOD) {
+		char kesPeriodString[50];
+		snprintf(kesPeriodString, SIZEOF(kesPeriodString), "%llu", ctx->kesPeriod);
```
janmazak, line 137: support for printf formats is limited on ledger; I think we only have %u and %d for uint32_t and int, respectively
so this won't work for integers not fitting into 32 bits and we need some more or less ugly hack (see e.g. what we do with tx ttl)

src/signOpCert.c
```diff
@@ -0,0 +1,169 @@
...
+		ui_displayPaginatedText(
+		        "KES period",
+		        kesPeriodString,
+		        this_fn
+		);
+	}
+	UI_STEP(UI_STEP_DISPLAY_ISSUE_COUNTER) {
+		char issueCounterString[50];
+		snprintf(issueCounterString, SIZEOF(issueCounterString), "%llu", ctx->kesPeriod);
```
janmazak, line 146: this should not be kesPeriod?

and the same problem as above

src/signOpCert.c
janmazak, line 55: Not `view.ptr`, but `&view`.

janmazak (commented):
Overall, I quite like it.

The only major problem is displaying uint64_t, again. 

Perhaps we should push ledger to fix it and support it properly in snprintf.

refi93:
the current version of the getPoolColdPublicKey call is returning the non-extended public key - I realized that even though it should technically be enough, it's likely to cause "pain" with the hw cli which internally deals with extended public key (especially in the hw signing file) and existing wallet implementations also likely use extended public keys, so this would only force them to introduce special code for cold keys.

TLDR - TODO rework getPoolColdPublicKey call to return the extended public key

src/signOpCert.c
refi93: in getPublicKeys we actually do ASSERT(FALSE):

https://github.com/vacuumlabs/ledger-app-cardano-shelley/blob/d031a6d18e988537e90792fd40a13294528ca40f/src/getPublicKeys.c#L53

But I agree ERR_NOT_IMPLEMENTED makes sense, so let's unify it

Makefile
```diff
@@ -29,7 +29,7 @@ PIN = 5555
 APPNAME = "Cardano ADA"
 APPVERSION = "2.1.0"
 
-APP_LOAD_PARAMS =--appFlags 0x240 --curve ed25519 --path "44'/1815'" --path "1852'/1815'"
+APP_LOAD_PARAMS =--appFlags 0x240 --curve ed25519 --path "44'/1815'" --path "1852'/1815'" --path "1853'/1815'"
```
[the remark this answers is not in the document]
refi93: I'd rather not introduce the flagging of features here as the solution is not final yet, so it is likely to result in conflicts. Maybe I would call the flag `HAVE_OPERATOR_FEATURES` to be consistent with similar flags for U2F (HAVE_U2F) and WebUSB (HAVE_WEBUSB)

src/bip44.c
```diff
@@ -158,6 +184,19 @@ bool bip44_isValidStakingKeyPath(const bip44_path_t* pathSpec)
...
 }
 
+// Pool cold keys
+bool bip44_isValidPoolColdKeyPath(const bip44_path_t* pathSpec)
+{
+	if (!bip44_hasValidCardanoPoolColdKeyPrefix(pathSpec)) return false;
+
+#define CHECK(cond) if (!(cond)) return false
+	CHECK(pathSpec->length == POOL_COLD_KEY_PATH_LENGTH);
+	CHECK(pathSpec->path[BIP44_I_POOL_COLD_KEY_USECASE] == 0);
```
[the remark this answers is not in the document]
refi93: Imho asserting in this case would make for a somewhat weird behavior as the concern of the function is to return true if it's a valid cold key path, otherwise false, so asserting for a certain subclass of paths based solely on their length does not feel right to me. I understand the concern that here we are potentially accessing something beyond the boundary, but it's obvious from the surrounding code that can't really happen

src/getPoolColdPublicKey.c
```diff
@@ -0,0 +1,106 @@
...
+static void getPoolColdPublicKey_ui_runStep();
+enum {
+	UI_STEP_WARNING = 100,
+	UI_STEP_DISPLAY_PATH,
+	UI_STEP_CONFIRM,
+	UI_STEP_RESPOND,
+	UI_STEP_INVALID,
+};
+
+void getPoolColdPublicKey_handleAPDU(
```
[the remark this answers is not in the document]
refi93: hmm, I'd think about it a bit more, I can imagine somebody having incentive to manage multiple pools, but I see it as a security/privacy feature having them export every single key separately, we may revisit that though in subsequent releases or ask about feedback when demonstrating the calls

src/bip44.h
```diff
@@ -33,12 +35,17 @@ enum {
...
+	BIP44_I_POOL_COLD_KEY_USECASE = 2,
+	BIP44_I_POOL_COLD_KEY = 3,
 };
 
 
 bool bip44_hasByronPrefix(const bip44_path_t* pathSpec);
 bool bip44_hasShelleyPrefix(const bip44_path_t* pathSpec);
-bool bip44_hasValidCardanoPrefix(const bip44_path_t* pathSpec);
+bool bip44_hasPoolColdKeyPrefix(const bip44_path_t* pathSpec);
+bool bip44_hasValidCardanoWalletPrefix(const bip44_path_t* pathSpec);
```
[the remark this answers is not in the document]
refi93: done, just github probably didn't pick it up for some reason

src/keyDerivation.c
```diff
@@ -18,7 +18,7 @@ void derivePrivateKey(
         privateKey_t* privateKey
 )
 {
-	if (!bip44_hasValidCardanoPrefix(pathSpec)) {
+	if (!bip44_hasValidCardanoWalletPrefix(pathSpec) && !bip44_isValidPoolColdKeyPath(pathSpec)) {
```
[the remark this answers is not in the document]
refi93: imho the current condition catches appropriately the "spirit" of what's going on for the reader, so I wouldn't abstract it away. Possibly the introduction of the compilation flag to separate operator app features would result in such refactor anyway

src/signOpCert.c
refi93: cold key path is a variable length entry - in theory I can have everything but the path in such struct and deal with the path separately but I'm not sure if it's worth it - what do you think?

src/cardano.h
```diff
@@ -16,7 +16,7 @@ STATIC_ASSERT(LOVELACE_MAX_SUPPLY < LOVELACE_INVALID, "bad LOVELACE_INVALID");
 #define VRF_KEY_HASH_LENGTH 32
 #define TX_HASH_LENGTH 32
 #define METADATA_HASH_LENGTH 32
-
+#define OP_CERT_BODY_LENGTH 48
```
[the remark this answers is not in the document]
refi93: done, github probably didn't pick it up

src/signOpCert.c
refi93: right, sorry I missed that while testing, added `str_formatUint64()` helper for that

src/signOpCert.c
refi93: with the uint64 values it gets a bit tricky as we don't have native functions to trace that - I can add it, though if not really needed at the moment, I would prefer not to

refi93: hmm, I added TRACE_UINT64, it may actually make debugging easier and I added tracing of such values here as well

refi93:
TODO - update cold key derivation to comply with CIP-1853: https://github.com/cardano-foundation/CIPs/pull/56 (i.e. change cold key usecase index to hardened)

janmazak:
Will be included as part of another PR.

PR #50, opened 2020-12-09
more secure memmove for read views

janmazak:
Useful to accidentally avoid reading more than a view contains (which os_memmove does not check).

### DIFF
--- a/src/bufView.h
+++ b/src/bufView.h
@@ -101,6 +101,15 @@ static inline cbor_token_t view_readToken(read_view_t* view)
 	return token;
 }
 
+// moves <length> bytes from the view to the buffer via os_memmove
+// (view.ptr is advanced accordingly)
+static inline void view_memmove(uint8_t* destBuffer, read_view_t* view, size_t length)
+{
+	ASSERT(length <= view_remainingSize(view));
+	os_memmove(destBuffer, view->ptr, length);
+	view_skipBytes(view, length);
+}
+
 // Note(ppershing): these macros expand to two arguments!
 #define VIEW_REMAINING_TO_TUPLE_BUF_SIZE(view) (view)->ptr, view_remainingSize(view)
 #define VIEW_PROCESSED_TO_TUPLE_BUF_SIZE(view) (view)->begin, view_processedSize(view)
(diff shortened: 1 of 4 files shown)


src/bufView.h
```diff
@@ -101,6 +101,16 @@ static inline cbor_token_t view_readToken(read_view_t* view)
 	return token;
 }
 
+// moves <length> bytes from the view to the buffer via os_memmove
+// (view.ptr is advanced accordingly)
+static inline void view_memmove(uint8_t* destBuffer, size_t destBufferSize, read_view_t* view, size_t length)
```
refi93, line 106: I find passing of the destBufferSize a bit artificial - for this function it's just a number that is compared and nothing else is actually done with it, it isn't really validated against the destBuffer and may set the caller up for false expectations that there is something more sophisticated about it, so I would rather leave that check to be done outside where the caller and reader has the actual context to the check in a meaningful way.

refi93 (changes requested):
see comment, otherwise LGTM

src/bufView.h
janmazak: The spirit of this code base is such that the caller of this would do the check anyway. 

IMHO the point of this function signature is that just the fact that the caller has to specify destBufferSize forces him to explicitly distinguish between destBufferSize and length (which is what we overlooked in the bug discovered by Ledger --- such an overlook would result in the same thing passed twice as argument to this function, and that would immediately stop me to think if that is really what I want to do).

refi93: If that's the concern, then it makes sense to make such wrapper around os_memmove as well, to retain a consistent interface for memmove operations and to make the caller explicitly specify and mentally check the destination buffer size there as well.

But I'm not really bought into that, to be honest - as the destBufferSize is effectively just a number that's being passed and then compared, and it even leaves space for free interpretation - another programmer may for example just see it as the "expected" destination size which should be the same as the size of the data being copied, and no automatic check would warn them against that. It would still be up to a human reviewer to catch that, but arguably, they would do the same without that parameter there. Introducing the parameter and using it wrong may even shift the focus away from the surrounding checks (or their lack) when reviewing/writing the code, leading to different bugs.

I would see value in it if the destination buffer size check was backed by some automated check, but this way it imho becomes just a mnemonic aid which may not really trigger a mental warning for other contributors and even if it did, the more you use it, the easier it is to just overlook it and it would no longer serve its purpose, as opposed to an automated check.

janmazak: OK, I can easily live without the argument :)

refi93 (approved):

PR #53, opened 2021-01-11
operator app

janmazak:
Contains an annoying bug: freezes ledger for pool registration txs.

Not complete yet (e.g. relays not displayed).


refi93:
note: we should also probably display the KES/VRF/COLD key following CIP-5: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0005/CIP-0005.md

janmazak:
Memory bugs are fixed now (afaik).

TODO:
(JM) doplnit test // TODO add pool retirement certificate
(JM) rebase nad pool-operator-app-refi93 (after #48 is finished and approved)
(JM) finish policyForSignTxCertificateStakePoolRetirement
(JM) finish policyForSignTxWitness
(JM) finish HANDLE_CERTIFICATE_POOL_RETIREMENT_STEP_DISPLAY_EPOCH // TODO; see also str_formatTtl()?
(JM) doplnit _parseCertificateData, ak treba // TODO no validation?
(JM) doriesit HANDLE_POOL_KEY_STEP_DISPLAY
(JM) spravit HANDLE_RELAY_STEP_DISPLAY
(JM) "we should also probably display the KES/VRF/COLD key following CIP-5: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0005/CIP-0005.md"
(JM) doplnit validaciu v signtxinit pre operatora? // TODO what other validations? // TODO think about this
(JM) precislovat usecasy od 1 (spolu s js kodom)
(JM) decision on whether to use POOL_OPERATOR_APP compilation flag (how much space does it save?), see also "TODO APPVERSION"

janmazak:
One unresolved issue in Makefile (might become a non-issue if we include all features unconditionally):

APP_LOAD_PARAMS =--appFlags 0x240 --curve ed25519 --path "44'/1815'" --path "1852'/1815'" --path "1853'/1815'"

This should be allowed only for the operator app, see the flag POOL_OPERATOR_APP.
Maybe I would call the flag HAVE_OPERATOR_FEATURES to be consistent with similar flags for U2F (HAVE_U2F) and WebUSB (HAVE_WEBUSB)

janmazak:
It seems to me everything is done, except validation and displaying DNS records in relays (especially SRV which could perhaps contain tabs).

Not tested properly yet --- not enough js integration tests yet, will be added soon.

Makefile
```diff
@@ -122,10 +133,12 @@ ifeq ($(GCCPATH),)
 $(info GCCPATH is not set: arm-none-eabi-* will be used from PATH)
 endif
 
+WERROR   := -Werror=incompatible-pointer-types -Werror=return-type
+
 CC       := $(CLANGPATH)clang
-CFLAGS   += -O3 -Os -Wall -Wextra -Wuninitialized
+CFLAGS   += -std=gnu11 -O3 -Os -Wall -Wextra -Wuninitialized $(WERROR)
```
refi93, line 139: why are these new flags needed?

src/bip44.c
```diff
@@ -158,7 +196,19 @@ bool bip44_isValidStakingKeyPath(const bip44_path_t* pathSpec)
...
 }
 
-// Futher
+#ifdef POOL_OPERATOR_APP
+bool bip44_isValidPoolColdKeyPath(const bip44_path_t* pathSpec)
+{
+#define CHECK(cond) if (!(cond)) return false
+	CHECK(pathSpec->length == BIP44_I_POOL_COLD_KEY + 1);
+	CHECK(bip44_hasValidCardanoPoolColdKeyPrefix(pathSpec));
+	CHECK(pathSpec->path[BIP44_I_POOL_COLD_KEY_USECASE] == 0);
```
refi93, line 205: this should be hardened as per the officially approved version of CIP-1853: https://github.com/cardano-foundation/CIPs/blob/master/CIP-1853/CIP-1853.md

src/signTxPoolRegistration.h
```diff
@@ -12,54 +12,84 @@
...
+	STAKE_POOL_REGISTRATION_CONFIRM = 2838,
+	STAKE_POOL_REGISTRATION_FINISHED = 2839,
 } sign_tx_pool_registration_state_t;
 
-enum {
-	SIGN_TX_POOL_OWNER_TYPE_PATH = 1,
-	SIGN_TX_POOL_OWNER_TYPE_KEY_HASH = 2,
-};
+typedef struct {
+	data_description_kind_t descriptionKind;
```
refi93, line 28: data description kind seems too generic to me - isn't this rather like "key reference type" - one being the key hash, the there key derivation path?

src/txHashBuilder.c
```diff
@@ -572,142 +673,150 @@ void txHashBuilder_addPoolRegistrationCertificate_enterRelays(tx_hash_builder_t*
...
-			} else {
-				BUILDER_APPEND_CBOR(CBOR_TYPE_NULL, 0);
-			}
+	if (ipv6->isNull) {
+		BUILDER_APPEND_CBOR(CBOR_TYPE_NULL, 0);
+	} else {
+		STATIC_ASSERT(SIZEOF(ipv6->ip) == IPV6_SIZE, "wrong ipv6 size");
+		BUILDER_APPEND_CBOR(CBOR_TYPE_BYTES, IPV6_SIZE);
+
+		// serialized as 4 big-endian uint32
```
refi93, line 719: why was that needed? wouldn't it be more straightforward to treat the ipv6 addresses as a byte array of 16 bytes all the time?

src/signTx.c
```diff
@@ -789,7 +830,72 @@ static void signTx_handleCertificate_ui_runStep()
...
+
+		ui_displayPaginatedText(
+		        "Retire stake pool",
+		        poolId,
+		        this_fn
+		);
+	}
+	UI_STEP(HANDLE_CERTIFICATE_POOL_RETIREMENT_STEP_DISPLAY_EPOCH) {
+		ui_displayUint64Screen(
+		        "Epoch",
```
refi93, line 872: I'd print a more user-friendly title clarifying when  should that retirement actually take effect, something like "Since end of epoch" - based on:
https://docs.cardano.org/projects/cardano-node/en/latest/stake-pool-operations/retire_stakepool.html

src/signTxPoolRegistration.c
```diff
@@ -101,235 +125,696 @@ static inline void advanceState()
...
 		if (subctx->currentOwner == subctx->numOwners) {
-			VALIDATE(subctx->numOwnersGivenByPath == 1, ERR_INVALID_DATA);
+			switch (commonTxData->signTxUsecase) {
+			case SIGN_TX_USECASE_POOL_REGISTRATION_OWNER:
+				VALIDATE(subctx->numOwnersGivenByPath == 1, ERR_INVALID_DATA);
+				break;
+
+				#ifdef POOL_OPERATOR_APP
+			case SIGN_TX_USECASE_POOL_REGISTRATION_OPERATOR:
+				VALIDATE(subctx->numOwnersGivenByPath <= 1, ERR_INVALID_DATA);
```
refi93, line 754: do we want to allow to the operator to sign as an owner at the same time?

src/signTx.c
```diff
@@ -383,14 +397,35 @@ static void signTx_handleInitAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
...
-			} else {
+				break;
+
+				#ifdef POOL_OPERATOR_APP
+			case SIGN_TX_USECASE_POOL_REGISTRATION_OPERATOR:
+				ASSERT(ctx->numCertificates == 1);
+				// inputs and withdrawals are unrestricted, to fund the tx
+				// only a single pool registration certificate
+				// with two possible witnesses: pool key and one owner
+				maxNumWitnesses = (size_t) ctx->numInputs +
```
refi93, line 412: imho there's not much point in allowing withdrawals for pool registration from operators' perspective if the owners would not be allowed to sign the transaction - technically, they can  use the cli, but I can see that inconsistency introducing significant confusion as to why operator's tx went through and owner's didn't

src/signTxPoolRegistration.c
refi93: it also seems to be in discrepancy with maxNumWitnesses which currently are taking into account only the amount of inputs and withdrawals (which I challenged if we actually need/want)

src/signTx.c
refi93: if operators want to withdraw rewards from their pool, they can do so simply by creating a separate withdrawal transaction without any certificates

src/signTxPoolRegistration.c
refi93: ok it isn't in discrepancy, I just noticed the additional "2" being summed into the equation

refi93: a possible problem I see with allowing signing a pool as an owner while in "operator mode" is that an attacker may unknowingly "coerce" an operator into signing up as an owner by requesting a witness for the operator's staking key which would be listed - the operator may or may not notice while checking the hashes that one of them is indeed their own - probably not a big deal but I would not mix the concerns unless really needed to prevent any such weird "loopholes"

src/signTx.c
```diff
@@ -365,10 +365,24 @@ static void signTx_handleInitAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
...
-		if (ctx->commonTxData.isSigningPoolRegistrationAsOwner) {
+		switch (ctx->commonTxData.signTxUsecase) {
+		case SIGN_TX_USECASE_POOL_REGISTRATION_OWNER:
+			// necessary for security reasons
 			VALIDATE(ctx->numCertificates == 1, ERR_INVALID_DATA);
 			VALIDATE(ctx->numWithdrawals == 0, ERR_INVALID_DATA);
+			break;
+
+			#ifdef POOL_OPERATOR_APP
+		case SIGN_TX_USECASE_POOL_REGISTRATION_OPERATOR:
```
refi93, line 376: imho we should enforce the same constraints for operators as for owners - any asymmetry in tx structure validation of those complementary usecases can cause unnecessary confusion and as a pool operator I'd appreciate a lot more the operation goes smoothly for everybody than being able to withdraw in the same transaction, even if I can do that completely separately

refi93 (changes requested):
see comments

src/signTxPoolRegistration.h
refi93: fair point and as long as "kind" is the standard suffix in the codebase I'm not against it, though it seems to me that it currently isn't, and switching ad-hoc to it may bring more confusion within the codebase when adding new features ("I see both instances of "type" and "kind" and I'm not sure how they differ" dilemma when naming new things) than the potential initial confusion of somebody coming from another codebase/language.

Also, what comes before that "type/kind" suffix should already be clear enough to be searchable on its own otherwise that can be a sign the name is too generic/misleading

src/uiScreens.c
```diff
@@ -461,17 +442,115 @@ void ui_displayPoolOwnerScreen(
...
+				        owner->keyHash, SIZEOF(owner->keyHash),
+				        rewardAddress, SIZEOF(rewardAddress)
+				);
+				break;
+			}
+			default:
+				ASSERT(false);
+			}
+
+			descLen += humanReadableAddress(
```
refi93, line 467: when the reward address is passed as a path, we should show that to the operator to give them a reasonable guarantee that they control the reward address and see what path it belongs to

src/getPoolColdPublicKey.c
```diff
@@ -0,0 +1,110 @@
+#ifdef POOL_OPERATOR_APP
```
refi93, line 1: general note - given that cold keys are indeed the same as standard public keys (perhaps only with a different policy) - I would consider discarding this dedicated call and merging it with getExtendedPublicKey, only adding a policy to display the cold key when deriving even during the bulk key export

src/textUtils.c
```diff
@@ -171,30 +171,56 @@ size_t str_formatMetadata(const uint8_t* metadataHash, size_t metadataHashSize,
...
-		if (buffer[i] <  33) return false;
+		if (buffer[i] <  32) return false;
 	}
 
 	return true;
 }
 
+bool str_isAllowedDnsName(const uint8_t* buffer, size_t bufferSize)
+{
+	if (str_isPrintableAsciiWithSpaces(buffer, bufferSize)) {
```
refi93, line 202: exiting early and returning false if there is some non-printable character would imho be a bit easier to follow

Makefile
```diff
@@ -26,13 +26,20 @@ NANOS_ID = 1
 WORDS = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about"
 PIN = 5555
 
+# pool operator app contains additional functionality
+# set to 1 if you want to include it, or to 0 if you want a regular app
+POOL_OPERATOR_APP = 1
```
refi93, line 31: we can already get rid of this as we ended up agreeing on not splitting the app (yet)

refi93: ok I see it's getting removed in the other PR

src/signTxPoolRegistration.c
```diff
@@ -101,235 +125,715 @@ static inline void advanceState()
...
+	TRACE_STACK_USAGE();
+	ui_callback_fn_t* this_fn = handlePoolRewardAccount_ui_runStep;
+
+	UI_STEP_BEGIN(subctx->ui_step, this_fn);
+
+	UI_STEP(HANDLE_POOL_REWARD_ACCOUNT_STEP_DISPLAY_PATH) {
+		ASSERT(subctx->stateData.poolRewardAccount.keyReferenceType == KEY_REFERENCE_PATH);
+
+		ui_displayPathScreen(
+		        "Reward account",
```
refi93, line 621: for the sake of consistency, I'd show the reward account in the same way as for ordinary users, i.e. its ordinal number, followed by the path

src/signTxPoolRegistration.c
```diff
@@ -406,16 +895,182 @@ static void signTxPoolRegistration_handleOwnerAPDU(uint8_t* wireDataBuffer, size
...
+	}
+	UI_STEP(HANDLE_RELAY_DNS_STEP_DISPLAY_DNSNAME) {
+		char dnsNameStr[1 + DNS_NAME_SIZE_MAX];
+		ASSERT(relay->dnsNameSize <= DNS_NAME_SIZE_MAX);
+		os_memcpy(dnsNameStr, relay->dnsName, relay->dnsNameSize);
+		dnsNameStr[relay->dnsNameSize] = '\0';
+		ASSERT(strlen(dnsNameStr) == relay->dnsNameSize);
+
+		ui_displayPaginatedText(
+		        "Dns name",
```
refi93, line 998:
```suggestion
		        "DNS name",
```

src/signTxPoolRegistration.c
```diff
@@ -406,16 +895,182 @@ static void signTxPoolRegistration_handleOwnerAPDU(uint8_t* wireDataBuffer, size
...
+	UI_STEP_BEGIN(subctx->ui_step, this_fn);
+
+	UI_STEP(HANDLE_RELAY_DNS_STEP_DISPLAY_NUMBER) {
+		ui_displayRelaycreen(
+			relay,
+			subctx->currentRelay,
+			this_fn
+		);
+	}
+	UI_STEP(HANDLE_RELAY_DNS_STEP_DISPLAY_PORT) {
```
refi93, line 979: It feels a bit weird to show the port first and then the DNS record name - I'd invert it

doc/ins_sign_stake_pool_registration.md
```diff
@@ -0,0 +1,191 @@
...
+
+- Transaction outputs must be given by addresses (not address params) and are not shown to the user (because he is not funding the transaction). The fee is not shown either.
+
+- Only a single witness is allowed (the key derivation path is the standard staking key path of the owner). 
+
+- It is necessary to include exactly one pool owner given by path in the certificate; all the other owners must be given by their respective key hashes.
+
+- Pool id must be given by key hash.
+
+- Pool reward account must be given in binary, not as a derivation path.
```
refi93, line 36: I think this requirement is superfluous - I see no harm in letting some of the owners own the rewards account

src/signTxPoolRegistration.c
```diff
@@ -101,235 +125,715 @@ static inline void advanceState()
...
-
-			TRACE_BUFFER(wireHeader->rewardAccount, SIZEOF(wireHeader->rewardAccount));
-			STATIC_ASSERT(SIZEOF(wireHeader->rewardAccount) == SIZEOF(p->rewardAccount), "wrong reward account size");
-			os_memmove(p->rewardAccount, wireHeader->rewardAccount, SIZEOF(p->rewardAccount));
-			const uint8_t header = getAddressHeader(p->rewardAccount, SIZEOF(p->rewardAccount));
-			VALIDATE(getAddressType(header) == REWARD, ERR_INVALID_DATA);
-			VALIDATE(getNetworkId(header) == commonTxData->networkId, ERR_INVALID_DATA);
-		}
+__noinline_due_to_stack__
+static void _calculateRewardAccount(
```
refi93, line 578: maybe I'd name it `_toRewardAccountBuffer`

src/signTxPoolRegistration.c
```diff
@@ -60,20 +65,39 @@ static inline void advanceState()
...
+	// intentional fallthrough
+
 	case STAKE_POOL_REGISTRATION_OWNERS:
 		ASSERT(subctx->currentOwner == subctx->numOwners);
-		ASSERT(subctx->numOwnersGivenByPath == 1);
 
+		txHashBuilder_addPoolRegistrationCertificate_enterRelays(txHashBuilder);
 		subctx->state = STAKE_POOL_REGISTRATION_RELAYS;
+
 		if (subctx->numRelays > 0) {
```
refi93, line 100: We should probably display to the operator that the pool has no relays, especially if we hide info about relays from owners, so it's easy to get confused

src/signTxPoolRegistration.c
refi93: same for `_calculatePoolKeyHash`

src/signTxPoolRegistration.c
```diff
@@ -750,18 +1333,46 @@ static void signTxPoolRegistration_handlePoolMetadataAPDU(uint8_t* wireDataBuffe
...
+			        "No pool owners",
+			        "",
+			        this_fn
+			);
+		} else {
+			UI_STEP_JUMP(HANDLE_CONFIRM_STEP_FINAL_NO_RELAYS);
+		}
+	}
+	UI_STEP(HANDLE_CONFIRM_STEP_FINAL_NO_RELAYS) {
+		if (subctx->numRelays == 0) {
```
refi93, line 1366: showing this in "owner mode" is confusing as we don't show the relays at all, if present

src/signOpCert.c
```diff
@@ -0,0 +1,177 @@
...
+	}
+	UI_STEP(UI_STEP_CONFIRM_START) {
+		ui_displayPrompt(
+		        "Start new",
+		        "operational certificate?",
+		        this_fn,
+		        respond_with_user_reject
+		);
+	}
+	UI_STEP(UI_STEP_DISPLAY_POOL_COLD_KEY_PATH) {
```
refi93, line 132: I'd align this with pool registration and show the bech32 pool id as well

src/signOpCert.c
```diff
@@ -0,0 +1,198 @@
...
+		ui_displayPathScreen("Pool cold key path", &ctx->poolColdKeyPathSpec, this_fn);
+	}
+	UI_STEP(UI_STEP_DISPLAY_POOL_ID) {
+		uint8_t poolKeyHash[POOL_KEY_HASH_LENGTH];
+		bip44_pathToKeyHash(&ctx->poolColdKeyPathSpec, poolKeyHash, SIZEOF(poolKeyHash));
+
+		ui_displayBech32Screen(
+		        "Pool ID",
+		        "pool_vk",
```
refi93, line 151: This should be `pool` as it's the pool key hash: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0005/CIP-0005.md#hashes

src/signTx.c
```diff
@@ -789,7 +825,70 @@ static void signTx_handleCertificate_ui_runStep()
...
+	ASSERT(ctx->stageData.certificate.type == CERTIFICATE_TYPE_STAKE_POOL_RETIREMENT);
+	ui_callback_fn_t* this_fn = signTx_handleCertificatePoolRetirement_ui_runStep;
+
+	UI_STEP_BEGIN(ctx->ui_step, this_fn);
+
+	UI_STEP(HANDLE_CERTIFICATE_POOL_RETIREMENT_STEP_DISPLAY_OPERATION) {
+		char poolId[200];
+		explicit_bzero(poolId, SIZEOF(poolId));
+
+		size_t length = encode_hex(
```
refi93, line 851: This should be the bech32 pool ID I presume, moreover, now that I'm testing it, I see

```
Retire stake pool
000000000000000000000
```
Do we really need to allocate 200 bytes here? Isn't some stack overflow going on here?

refi93: I think that the pool key hash fetched from the instruction state is actually empty as it is computed into a local variable when loading it into the tx hash builder:

https://github.com/vacuumlabs/ledger-app-cardano-shelley/blob/01ecd80202758eda1d64901958811feef36081da/src/signTx.c#L984

but then it's fetched from the "poolKeyHash" property which is not passed as part of the certificate data:

https://github.com/vacuumlabs/ledger-app-cardano-shelley/blob/01ecd80202758eda1d64901958811feef36081da/src/signTx.c#L852

refi93:
@janmazak Found one more inconsistency related to stake pool id -> in `signTx_handleCertificate_ui_runStep()` we are still displaying the pool id as a hexstring instead of its bech32 representation

src/bip44.c
```diff
@@ -212,10 +247,100 @@ size_t bip44_printToStr(const bip44_path_t* pathSpec, char* out, size_t outSize)
...
+				return PATH_INVALID;
+			}
+		}
+
+		default:
+			return PATH_INVALID;
+		}
+	}
+
+	return PATH_INVALID;
```
mkv-vcm, line 293: Is there some common coding ruleset for the company or the project I can read about? These four returns (instead of just the one at the end) might improve or worsen readability, depending on the stance of the people involved, and I'm curious if there's a written one.

janmazak: I don't think we have such a ruleset for hardware wallets or Adalite (and definitely not one company-wide, at least not one I know of; there are many projects and they are quite diverse, but I believe some bigger teams might have some rules). We use some code formatters, though (eslint for js, astyle for C).

In this case, there are many reasons why a path is invalid, and new ones are popping up every other while as new valid path schemes are added. This particular method will be refactored (split into several parts at least) soon when multisig is added, i.e. paths starting with 1854' instead of 1852' (https://github.com/cardano-foundation/CIPs/blob/master/CIP-1854/CIP-1854.md). Code improvement suggestions are always welcome, though. I will think about whether do some changes right away.

janmazak:
Merged at Ledger repo.

PR #54, opened 2021-01-28
Multiassets

janmazak:
Not 100% finished and needs some polishing.


src/signTx.c
```diff
@@ -286,26 +323,52 @@ static void signTx_handleInitAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
 		ctx->commonTxData.protocolMagic = u4be_read(wireHeader->protocolMagic);
 		TRACE("protocol magic %d", ctx->commonTxData.protocolMagic);
 
+		switch (wireHeader->includeTtl) {
+		case SIGN_TX_INCLUDED_YES:
```
refi93, line 327: may be worth factoring out the logic to parse the YES/NO enum into a boolean value, given that it's repeated three times already

src/signTx.c
```diff
@@ -781,6 +664,7 @@ static void signTx_handleTtl_ui_runStep()
 
 	UI_STEP(HANDLE_TTL_STEP_DISPLAY) {
 		char ttlString[50];
+		// TODO just display it as a plain uint64, the same as validity interval start
```
refi93, line 667: as we discussed yesterday, I'd keep the TTL formatting logic and perhaps generalize the name to `str_formatValidityBoundary` or something like that

src/signTx.c
```diff
@@ -1262,7 +1149,9 @@ static void signTx_handleMetadataAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_
 		TRACE_BUFFER(wireDataBuffer, wireDataSize);
 
 		VALIDATE(wireDataSize == METADATA_HASH_LENGTH, ERR_INVALID_DATA);
+		STATIC_ASSERT(SIZEOF(ctx->stageData.metadata.metadataHash) == METADATA_HASH_LENGTH, "inconsistent metadata hash length");
```
refi93, line 1152: I'd say "wrong" rather than "inconsistent"

src/signTx.c
```diff
@@ -1294,10 +1183,87 @@ static void signTx_handleMetadataAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_
...
+		);
+	}
+	UI_STEP(HANDLE_VALIDITY_INTERVAL_STEP_RESPOND) {
+		respondSuccessEmptyMsg();
+		advanceStage();
+	}
+	UI_STEP_END(HANDLE_VALIDITY_INTERVAL_STEP_INVALID);
+}
+
+static void signTx_handleValidityIntervalAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wireDataSize)
```
refi93, line 1215: I'd rather name it `signTx_handleValidityIntervalStartAPDU` given that `validityInterval` alone can mislead the reader into thinking that both validityIntervalStart and TTL are being handled here

src/signTx.h
```diff
@@ -69,43 +67,46 @@ typedef struct {
...
+	uint16_t numWitnesses;
 
 	uint16_t currentInput;
 	uint16_t currentOutput;
 	uint16_t currentCertificate;
 	uint16_t currentWithdrawal;
 	uint16_t currentWitness;
 
+// TODO add js tests for missing ttl and inclusion of validity interval start
+// TODO display ttl with pershing routine, also validity interval start, but only on mainnet?
```
refi93, line 86: That is sensible IMHO

src/signTxOutput.c
```diff
@@ -0,0 +1,664 @@
...
+		ASSERT(subctx->numTokenAmounts > 0);
+
+		subctx->state = STATE_OUTPUT_TOKEN_AMOUNT;
+		break;
+
+	case STATE_OUTPUT_TOKEN_AMOUNT:
+		// we are done with the current token group
+		ASSERT(subctx->currentTokenAmount == subctx->numTokenAmounts);
+		ASSERT(subctx->currentTokenGroup < subctx->numTokenGroups);
+		subctx->currentTokenGroup++;
```
refi93, line 83: shouldn't this reset `currentTokenAmount` to zero?

src/signTxOutput.c
```diff
@@ -0,0 +1,664 @@
...
+
+		// we are going to receive token amounts for this group
+		ASSERT(subctx->numTokenAmounts > 0);
+
+		subctx->state = STATE_OUTPUT_TOKEN_AMOUNT;
+		break;
+
+	case STATE_OUTPUT_TOKEN_AMOUNT:
+		// we are done with the current token group
+		ASSERT(subctx->currentTokenAmount == subctx->numTokenAmounts);
```
refi93, line 81: this sees a bit weird to me at first sight - I would expect the state machine to enter this state at the beginning of processing the sequence of token amounts for the respective token group - why isn't it that straightforward?

src/signTxOutput.c
```diff
@@ -0,0 +1,664 @@
...
+			subctx->state = STATE_OUTPUT_CONFIRM;
+		}
+		break;
+
+	case STATE_OUTPUT_TOKEN_GROUP:
+		ASSERT(subctx->currentTokenGroup < subctx->numTokenGroups);
+
+		// we are going to receive token amounts for this group
+		ASSERT(subctx->numTokenAmounts > 0);
+
```
refi93, line 75: I guess it makes sense to assert on currentTokenAmount being zero here

src/signTxOutput.c
```diff
@@ -0,0 +1,664 @@
...
+		token_amount_t* tokenAmount = &subctx->stateData.tokenAmount;
+		if (str_isTextPrintable(tokenAmount->assetName, tokenAmount->assetNameSize)) {
+			char name[ASSET_NAME_SIZE_MAX + 1];
+			ASSERT(tokenAmount->assetNameSize + 1 <= SIZEOF(name));
+			for (size_t i = 0; i < tokenAmount->assetNameSize; i++)
+				name[i] = tokenAmount->assetName[i];
+			name[tokenAmount->assetNameSize] = '\0';
+
+			ui_displayPaginatedText(
+			        "Token name",
```
refi93, line 461: I'd display `Token name (ASCII)` or `Token name (text)` otherwise it opens up space for confusion should the token's name happen to be a valid hex string

src/signTxOutput.c
```diff
@@ -0,0 +1,664 @@
...
+		VALIDATE(tokenAmount->assetNameSize > 0, ERR_INVALID_DATA);
+		VALIDATE(tokenAmount->assetNameSize <= ASSET_NAME_SIZE_MAX, ERR_INVALID_DATA);
+
+		ASSERT(tokenAmount->assetNameSize <= SIZEOF(tokenAmount->assetName));
+		view_memmove(tokenAmount->assetName, &view, tokenAmount->assetNameSize);
+
+		VALIDATE(view_remainingSize(&view) == 8, ERR_INVALID_DATA);
+		tokenAmount->amount = parse_u8be(&view);
+		TRACE_UINT64(tokenAmount->amount);
+		// TODO validate something?
```
refi93, line 519: did you have something specific in mind? If not I think we don't really have anything to validate as we don't have further context about the specific token

src/textUtils.c
```diff
@@ -139,6 +182,19 @@ void str_validateTextBuffer(const uint8_t* text, size_t textSize)
 	}
 }
 
+// check if it is printable ASCII between 33 and 126
+bool str_isTextPrintable(const uint8_t* text, size_t textSize)
```
refi93, line 186: I find it a bit unsettling that this function would not work properly for null-terminated strings, not sure though what I would do differently - maybe rename it to `str_isAsciiPrintableBuffer()` to convey that the function does not really expect a proper C string

src/txHashBuilder.c
```diff
@@ -68,6 +79,11 @@ void txHashBuilder_init(
 		builder->includeMetadata = includeMetadata;
 		if (includeMetadata) numItems++;
 
+		builder->includeValidityIntervalStart = includeValidityIntervalStart;
+		if (includeValidityIntervalStart) numItems++;
+
+		ASSERT(3 <= numItems && numItems <= 8);
```
refi93, line 85: I'd wrap the expressions in parentheses for peace of mind

refi93 (changes requested):
see comments

src/uiScreens.c
```diff
@@ -172,7 +173,34 @@ void ui_displayStakingInfoScreen(
...
+	if (str_isAsciiPrintableBuffer(token->assetName, token->assetNameSize)) {
+		char name[ASSET_NAME_SIZE_MAX + 1];
+		ASSERT(token->assetNameSize + 1 <= SIZEOF(name));
+		os_memmove(name, token->assetName, token->assetNameSize);
+		name[token->assetNameSize] = '\0';
+
+		bool isEmpty = (token->assetNameSize == 0);
+
+		ui_displayPaginatedText(
+		        (isEmpty) ? "Token name (empty):" : "Token name (ASCII):",
```
refi93, line 190:
```suggestion
		        (isEmpty) ? "Token name is empty" : "Token name (ASCII):",
```

refi93: also, in Trezor we use the term "asset name" which also aligns with the terminology used in developer docs: https://developers.cardano.org/en/development-environments/native-tokens/multi-asset-tokens-explainer/ so I'd prefer showing "asset name" over "token name"

src/txHashBuilder_test.c
```diff
@@ -330,13 +336,16 @@ void run_txHashBuilder_test()
...
+		txHashBuilder_addOutput_basicData(
 		        &builder,
 		        tmp, tmpSize,
-		        it->amount
+		        it->amount,
+		        0
 		);
 	}
 
+	// TODO multiasset output? covered in integration tests
```
refi93, line 347: what's the issue with adding a sample multiasset output here for the sake of completeness?

src/txHashBuilder.c
```diff
@@ -7,6 +7,18 @@
 #include "crc32.h"
 #include "bufView.h"
 
+// this tracing is rarely needed
+// so we want to keep it turned off to avoid polluting the trace log
+
+//#define TRACE_TX_HASH_BUILDER
+
+#ifdef TRACE_TX_HASH_BUILDER
```
refi93, line 15: what's the benefit of such tracing? Even if it was there I wouldn't really be able to tell at first sight what's it for so I'd most likely skip it even if was actually beneficial

src/addressUtilsShelley.c
```diff
@@ -504,6 +502,4 @@ void parseAddressParams(const uint8_t *wireDataBuffer, size_t wireDataSize, addr
 	default:
 		ASSERT(false);
 	}
-
-	VALIDATE(view_remainingSize(&view) == 0, ERR_INVALID_DATA);
```
refi93, line 508: why the check that the view was fully consumed is no longer enforced?

src/cardanoOutputs.h
```diff
@@ -0,0 +1,18 @@
...
+
+#include "cardano.h"
+
+
+typedef struct {
+	uint8_t policyId[MINTING_POLICY_ID_SIZE];
+} token_group_t;
+
+typedef struct {
+	uint8_t assetName[ASSET_NAME_SIZE_MAX];
```
refi93, line 12: assetNameBytes is imho a better name to not mislead anybody into thinking it's a (null-terminated) string

src/securityPolicy.c
```diff
@@ -266,6 +266,39 @@ security_policy_t policyForSignTxOutputAddressParams(
...
 
+// multiasset token groups in outputs
+security_policy_t policyForSignTxOutputAssetGroup(
+        security_policy_t addressPolicy,
+        token_group_t* tokenGroup MARK_UNUSED
+)
+{
+	ALLOW_IF(addressPolicy == POLICY_ALLOW_WITHOUT_PROMPT);
+
+	SHOW_IF(true); // TODO replace with SHOW(); in operator app
```
refi93, line 277: seems like an easy win to replace it now

src/signTx.c
```diff
@@ -63,16 +56,23 @@ static inline void advanceStage()
...
+			// wait for TTL APDU
+			break;
+		}
+
+	// intentional fallthrough
 
 	case SIGN_STAGE_TTL:
-		// we should have received ttl
-		ASSERT(ctx->ttlReceived);
+		if (ctx->includeTtl)
```
refi93, line 71:
```suggestion
		if (ctx->includeTtl) {
		    ASSERT(ctx->ttlReceived);
		}
```

refi93, line 71: I'd wrap it in curly braces, imho the one saved line is not worth it

src/signTxOutput.c
```diff
@@ -0,0 +1,651 @@
...
+#	undef   CASE
+		default:
+			THROW(ERR_NOT_IMPLEMENTED);
+		}
+
+		signTx_handleOutput_addressParams_ui_runStep();
+	}
+}
+
+static void signTxOutput_handleBasicDataAPDU(uint8_t* wireDataBuffer, size_t wireDataSize)
```
refi93, line 264: "basic" is kind of an umbrella term, IMHO a name like `signTxOutput_handleTopLevelDataAPDU` would do a better job conveying the essence of what's going on here

src/txHashBuilder.c
```diff
@@ -133,34 +169,139 @@ void txHashBuilder_enterOutputs(tx_hash_builder_t* builder)
...
+		//     ]
+		//   ]
+		// ]
+		{
+			BUILDER_APPEND_CBOR(CBOR_TYPE_ARRAY, 2);
+			{
+				BUILDER_APPEND_CBOR(CBOR_TYPE_BYTES, addressSize);
+				BUILDER_APPEND_DATA(addressBuffer, addressSize);
+			}
+			{
```
refi93, line 218: conceptually, shouldn't the `BUILDER_APPEND_CBOR(CBOR_TYPE_ARRAY, 2);` be besides the previous two APPEND statements, hence within the same scope?

src/txHashBuilder.c
```diff
@@ -658,11 +807,14 @@ void txHashBuilder_addWithdrawal(
...
 )
 {
+	_TRACE("state = %u, remainingWithdrawals = %u", builder->state, builder->remainingWithdrawals);
+
 	ASSERT(builder->state == TX_HASH_BUILDER_IN_WITHDRAWALS);
 	ASSERT(builder->remainingWithdrawals > 0);
-	ASSERT(rewardAddressSize == 1 + ADDRESS_KEY_HASH_LENGTH);
 	builder->remainingWithdrawals--;
 
+	ASSERT(rewardAddressSize == 1 + ADDRESS_KEY_HASH_LENGTH);
```
refi93, line 816: seems like a redundant change

src/txHashBuilder.c
```diff
@@ -677,14 +829,15 @@ void txHashBuilder_addWithdrawal(
...
 static void txHashBuilder_assertCanLeaveWithdrawals(tx_hash_builder_t* builder)
 {
+	_TRACE("state = %u, remainingWithdrawals = %u", builder->state, builder->remainingWithdrawals);
+
 	ASSERT(builder->state <= TX_HASH_BUILDER_IN_WITHDRAWALS);
 
-	if (builder->state == TX_HASH_BUILDER_IN_WITHDRAWALS) {
-		ASSERT(builder->remainingWithdrawals == 0);
-	} else {
+	if (builder->state < TX_HASH_BUILDER_IN_WITHDRAWALS) {
```
refi93, line 836: doing a `<` comparison against an enum feels kind of wrong and can lead to pretty weird bugs if new enums not necessarily respecting the ordering are introduced later on - wouldn't `!=` have the same effect?

refi93 (changes requested):
see comments

src/txHashBuilder.c
janmazak: You probably never tried to debug tx hash builder :) It has its own state machine, so most changes in the sign tx state machine are reflected for tx hash builder, too. If you mess it up, as I did this time, tracing is very handy, and without this, you would have to write it ad-hoc (and waste some time by repeatedly deploying code to ledger which gets stuck every time, just to get the tracing right).

On the other hand, I don't want to have the tracing active all the time, because it pollutes the log (data are added to the tx hash builder often in small bits). Usually the tx hash builder is easy to fix, time devoted to that is, say, 5-10% of the total time spent on modifying signTx.

src/addressUtilsShelley.c
janmazak: It should never have been here in the first place; it's just that before now, address params were always at the end of wireData, so the view contents was consumed in every usecase. This is no longer true; and the caller should be responsible for determining if more data are to be read from the view or not.

src/txHashBuilder.c
janmazak: Each of the two elements of the top-level array has its own scope. The scope is not used to encompass all array elements (that's the role of the outermost scope).

src/txHashBuilder.c
janmazak: After this change, argument checking is separated from builder state checking. So the logic is "trace the state of the builder, assert the state is consistent and that we are supposed to call this function at all, assert that the given arguments are correct etc.".

src/txHashBuilder.c
janmazak: I was wary of inequalities here, but Pershing introduced them, and they make sense for a linear ordering. The alternative is to use a (potentially long) list of preceding states in a switch statement, and I don't like that much, since it would have to be modified if a new state is introduced in the linear ordering (not much likely to happen, though). 

If we make <= comparison (just above), then I find < also viable. Using != gives less information (one has to read it together with <= to actually know what the if condition describes).

Since the time the inequalities were introduced, I have added "substates" (for pool registration certificates and outputs), so the ordering is not as linear as it was before. I guess I should rewrite it into the explicit switch statement to be on the safe side.

src/uiScreens.c
```diff
@@ -191,6 +219,48 @@ void ui_displayAmountScreen(
...
+        const char* screenHeader,
+        uint64_t boundary,
+        uint8_t networkId, uint32_t protocolMagic,
+        ui_callback_fn_t callback
+)
+{
+	char boundaryStr[30];
+	explicit_bzero(boundaryStr, SIZEOF(boundaryStr));
+
+	if ((networkId == MAINNET_NETWORK_ID) && (protocolMagic == MAINNET_PROTOCOL_MAGIC)) {
```
refi93, line 248: this probably deserves an explanatory comment that we know the epoch/slot configuration reliably only for mainnet

src/textUtils.h
```diff
@@ -15,11 +17,22 @@ void str_traceAdaAmount(const char *prefix, uint64_t amount);
...
+void str_traceUint64(uint64_t number);
+#define TRACE_UINT64(NUMBER) \
+	do { \
+		str_traceUint64(NUMBER); \
+	} while(0)
+#else
+#define TRACE_UINT64(NUMBER)
+#endif
+
+size_t str_formatValidityBoundary(uint64_t ttl, char* out, size_t outSize);
```
refi93, line 30:
```suggestion
size_t str_formatValidityBoundary(uint64_t slotNumber, char* out, size_t outSize);
```

src/signTx.c
```diff
@@ -1294,10 +1151,88 @@ static void signTx_handleMetadataAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_
...
 
+// ============================== VALIDITY INTERVAL START ==============================
+
+enum {
+	HANDLE_VALIDITY_INTERVAL_START_STEP_DISPLAY = 900,
+	HANDLE_VALIDITY_INTERVAL_START_STEP_RESPOND,
+	HANDLE_VALIDITY_INTERVAL_START_STEP_INVALID,
+};
+
+static void signTx_handleValidityInterval_ui_runStep()
```
refi93, line 1162: UX-wise I find it a bit counterintuitive that validity interval start is being confirmed after TTL, but I guess reversing the order is not really an option given that it's coupled with tx serialization :/

Also, it may make sense to explicitly notify the user that the tx has no ttl/validity start

I'm wondering if it would not make sense to (eventually) move ttl/validity start directly to the init APDU to make the UX around those values more intuitive and remove that "machinery" just for two numbers

src/txHashBuilder_test.c
janmazak: I have added it. I was afraid it might consume too much memory, but I have reused buffers and refactored it a tiny bit and no problem appeared.

src/signTx.c
janmazak: I don't understand why they did not just replace ttl with validity interval (in the same way as coin in output is replaced by coin / [coin, multiasset]). That would fix all ui problems and also look more natural.

On one hand, we don't have much time to do changes; on the other, why show the users the "wrong" user flow with validity interval if we can avoid it?...

refi93 (approved):

refi93:
@janmazak don't forget to update the docs: https://github.com/vacuumlabs/ledger-app-cardano-shelley/blob/develop/doc/ins_sign_tx.md so they don't fall out of sync

PR #57, opened 2021-02-11
Multiasset fix security policy for tokens

janmazak:


src/securityPolicy.c
```diff
@@ -293,21 +302,39 @@ security_policy_t policyForSignTxOutputToken(
...
+		SHOW();
+
+	default:
+		ASSERT(false);
+	}
+
+	DENY(); // should not be reached
 }
 
 security_policy_t policyForSignTxOutputConfirm(
```
refi93, line 319: Is there any point in separating the policies for different output entries if they are basically the same? I can see them getting unintentionally misaligned if we opt to change outputs and general and forget about tokens or vice-versa.

So I would opt for a single "policyForSignTxOutputEntries" which seems like what we indeed want to achieve there and split it only once we would actually need to.

src/securityPolicy.c
```diff
@@ -282,9 +282,18 @@ security_policy_t policyForSignTxOutputAssetGroup(
         token_group_t* tokenGroup MARK_UNUSED
```
refi93, line 282: is there any point in having this function which just mirrors the `addressPolicy`? Why not directly using `addressPolicy` within the function to handle asset groups/tokens to make it clear they should share the same policy?

src/signTxOutput.c
```diff
@@ -226,6 +226,8 @@ static void signTx_handleOutput_addressParams()
 	TRACE("Policy: %d", (int) policy);
 	ENSURE_NOT_DENIED(policy);
 
+	subctx->addressSecurityPolicy = policy;
```
refi93, line 229: If I understand correctly, the `subctx` is initialized in `signTxOutput_init()` which sets all fields to zero, including the addressSecurityPolicy, which is an invalid value. Do you know how is it possible that Ledger did not reject the transaction?

src/securityPolicy.c
refi93: I would perhaps rename the `subctx->addressSecurityPolicy` to `subctx->outputSecurityPolicy` to reflect that if you agree with my suggestion

refi93 (changes requested):
few improvement/simplification suggestions

src/signTxOutput.c
refi93: ah got it - the policies were written in a way that the zero value feel through to `SHOW()` :facepalm:

refi93: or shown the tokens anyway, looking at the original logic of `policyForSignTxOutputAssetGroup`

src/securityPolicy.c
refi93: and same comment for `policyForSignTxOutputToken`

src/signTxOutput.c
refi93: ah sorry, I realized the problem was indeed that the tokens defaulted to be shown, I thought the opposite was happening for some reason during the review and got confused

src/securityPolicy.c
```diff
@@ -293,21 +302,39 @@ security_policy_t policyForSignTxOutputToken(
...
+	case POLICY_SHOW_BEFORE_RESPONSE:
+		PROMPT_IF(numAssetGroups > 0);
+		ALLOW();
+
+	default:
+		ASSERT(false);
+	}
 
-	PROMPT_IF(numAssetGroups > 0);
+	DENY(); // should not be reached
```
refi93, line 336: the comment is imho too self evident and doesn't add meaningful insight. We already do this in other part of the policies without explaining it

src/securityPolicy.c
```diff
@@ -382,11 +409,9 @@ security_policy_t policyForSignTxStakePoolRegistrationOwner(pool_owner_t* owner)
 	switch (owner->ownerType) {
 	case SIGN_TX_POOL_OWNER_TYPE_KEY_HASH:
 		SHOW();
```
refi93, line 411: why removing the break?

refi93 (approved):

PR #58, opened 2021-02-15
Refactor endif and fix include guards

janmazak:


refi93 (approved):
LGTM, seems like there is ongoing work in vscode's c/c++ plugin to add support for navigating through matching pairs of those directives: https://github.com/microsoft/vscode-cpptools/issues/4779 which should make our lives easier

janmazak:
Will be included as part of another PR.

PR #59, opened 2021-02-26
operator app doc

janmazak:


CHANGELOG.md
```diff
@@ -1,11 +1,33 @@
 # Change Log
+
 All notable changes to this project will be documented in this file.
  
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [2.2.0](https://github.com/LedgerHQ/app-cardano/compare/2.1.0...LedgerHQ:2.2.0) - [TBD]
```
refi93, line 8: should be bumped - I guess we are still deciding whether to do a major version or minor version bump - I would personally go with minor here. The previous changes were also "breaking" in the sense that the init apdu was being modified, but feature-wise the changes were just additions which is the case here as well and other flows apart from signing stake pool registration shouldn't be affected, if I understand correctly

janmazak:
Merged into operatorapp.

PR #60, opened 2021-03-02
WIP: Unify public key export

janmazak:
Cold keys and wallet keys in a single call.


src/bip44.h
```diff
@@ -64,17 +62,41 @@ bool bip44_isValidStakingKeyPath(const bip44_path_t* pathSpec);
...
 
 size_t bip44_printToStr(const bip44_path_t*, char* out, size_t outSize);
 
 
+typedef enum {
+	// hd wallet account
+	PATH_WALLET_ACCOUNT,
+
+	// hd wallet address
+	PATH_WALLET_ADDRESS,
```
refi93, line 79: PATH_WALLET_SPENDING_KEY would be less ambiguous of a name, as technically, the staking key is also included in addresses

src/bip44.c
```diff
@@ -260,6 +245,76 @@ size_t bip44_printToStr(const bip44_path_t* pathSpec, char* out, size_t outSize)
...
 }
 
+
+bip44_path_type_t bip44_classifyPath(const bip44_path_t* pathSpec)
+{
+	if (bip44_hasValidCardanoPoolColdKeyPrefix(pathSpec)) {
+		if (bip44_isValidPoolColdKeyPath(pathSpec)) {
+			return PATH_POOL_COLD_KEY;
+		} else {
+			return PATH_POOL_COLD_KEY_INVALID;
```
refi93, line 255: I don't see any code actually dealing with invalid paths on this granularity - I would just default to `PATH_INVALID` (renaming the PATH_UNKNOWN enum, so it's more accurate)

We can always introduce that kind of granularity if we would actually need it, but for now it seems like unnecessary bloat making this logic more complicated than it needs to be, unless I'm missing something

src/bip44.c
```diff
@@ -260,6 +245,76 @@ size_t bip44_printToStr(const bip44_path_t* pathSpec, char* out, size_t outSize)
...
+
+		case 3: {
+			return PATH_WALLET_ACCOUNT;
+		}
+
+		case 5: {
+			const uint8_t chainType = bip44_getChainTypeValue(pathSpec);
+			switch (chainType) {
+
+			case 0:
```
refi93, line 270: don't we have proper enums for chain type?

src/securityPolicy.c
```diff
@@ -624,41 +578,67 @@ security_policy_t policyForSignTxWitness(
...
+		switch (bip44_classifyPath(pathSpec)) {
 
-		} else {
+		case PATH_WALLET_ADDRESS:
+		case PATH_WALLET_STAKING_KEY:
+		case PATH_POOL_COLD_KEY:
+			if (bip44_isPathReasonable(pathSpec)) {
+				ALLOW();
+			} else {
+				WARN();
```
refi93, line 590: this seems to be more strict than the original logic which allows any valid key - I find it a pointless to warn for unreasonable paths, especially if we don't return the public key - if the tx contains entries actually defined by such keys we already warn the user si warning them again during signing seems kind of redundant to me

refi93: ah I see, the "WARN()" directive was embedded in the functions, so it's not really different, maybe apart from cold keys

janmazak: The problem with witnesses is that they can be totally unrelated to the tx (no way to check). So whenever something suspicious occurs, it is better to warn the user.

janmazak:
Merged into operatorapp.

PR #61, opened 2021-03-10
Asset fingerprint

janmazak:


src/hash_test.c
```diff
@@ -44,6 +44,21 @@ void testcase_blake2b_224(const char* inputHex, const char* expectedHex)
 	EXPECT_EQ_BYTES(expectedBuffer, outputBuffer, expectedSize);
 }
 
+void testcase_blake2b_160(const char* inputHex, const char* expectedHex)
```
refi93, line 47: I'd also add a test to verify that the asset fingerprints produced are up to specs

src/uiScreens.c
```diff
@@ -173,31 +173,33 @@ void ui_displayStakingInfoScreen(
...
-		        callback
-		);
-	} else {
-		ui_displayHexBufferScreen(
-		        "Asset name (hex):",
-		        token->assetNameBytes, token->assetNameSize,
-		        callback
-		);
-	}
+	ui_displayBech32Screen(
```
refi93, line 197: I'd rather not couple bech32 encoding with screen rendering logic. Also, I don't see this function being reused for other contexts where it would make sense like showing pool id or vrf key in the pool registration certificate

janmazak: It is reused in pool operator app, I actually extracted it from there.

janmazak: This function (similarly to hexBufferScreen etc.) provides a crucial feature that is not obvious: it allocates the buffer for the bech32 string locally. If I have a ui handler with, say, 5 steps, and each of them needs a string buffer, I cannot afford to allocate all of the buffers in the handler because there is not enough stack space. So a separate function, marked __noinline_due_to_stack__, must be called (see the operator app PR). And it is exactly what this function does: encodes a buffer into bech32 and displays it. I don't see any way around this if we are only using static memory.

src/uiScreens.c
```diff
@@ -173,31 +173,58 @@ void ui_displayStakingInfoScreen(
...
-		);
-	} else {
-		ui_displayHexBufferScreen(
-		        "Asset name (hex):",
-		        token->assetNameBytes, token->assetNameSize,
-		        callback
-		);
-	}
+	ui_displayPaginatedText(
+	        "Token asset",
```
refi93, line 224: `Asset fingerprint:` is what we print in Trezor: https://github.com/trezor/trezor-firmware/pull/1510/files#diff-718c35fab848e1ee6f5e2a387c71b0a0ea8d84039fedce530cb9f0584c164495R104

If it doesn't fit, I'd prefer "Asset ID", is the "Token asset" some coined term you came across?

src/uiScreens.c
```diff
@@ -173,31 +173,58 @@ void ui_displayStakingInfoScreen(
...
+	return len;
+}
+
+void ui_displayAssetFingerprintScreen(
+        token_group_t* tokenGroup,
         token_amount_t* token,
         ui_callback_fn_t callback
 )
 {
-	if (str_isAsciiPrintableBuffer(token->assetNameBytes, token->assetNameSize)) {
```
refi93, line 181: I think `str_isAsciiPrintableBuffer` function is no longer used

refi93 (approved):

refi93 (changes requested):
ok I noticed when reviewing the UI that "Asset fingerprint" prompt is followed by "Token amount" - I can see this being confusing for users so I'd consider rewording the "Token amount" prompt to "Asset amount sent" - to not expose users to any ambiguity

refi93 (approved):
ok let's keep Token amount, it should be fairly obvious from the flow that it's related to the asset fingerprint one screen earlier

PR #63, opened 2021-03-17
Catalyst registration support

refi93:
Motivation: Add support for Catalyst voting registration

Changes:
* renamed metadata to auxData in the code to reflect latest CDDL
* introduce new hash builder for auxiliary data (former metadata in pre-Mary era) needed to compute the hash of the auxiliary data as such as well as the catalyst registration payload (needed to obtain the signature)
* introduce new dedicated submachine for Catalyst voting registration which triggers from the Auxiliary data step within the main signTx flow
* the catalyst registration submachine, once in its terminal state, returns the overall auxiliary data hash and registration signature in a single APDU

Testing:
* added device test for auxiliary data hash builde
* ran `yarn test-integration --grep signTxOrdinaryMary` on https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/99 with a Catalyst testcase added

TODOs:
* rebase on top of current develop, currently it's based on top of operator app (at least we know the two features can coexist)


src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,573 @@
...
+
+	UI_STEP(HANDLE_VOTING_REWARDS_ADDRESS_PARAMS_STEP_WARNING) {
+		ui_displayPaginatedText(
+		        "Unusual request",
+		        "Proceed with care",
+		        this_fn
+		);
+	}
+	UI_STEP(HANDLE_VOTING_REWARDS_ADDRESS_PARAMS_STEP_DISPLAY_ADDRESS) {
+		uint8_t addressBuffer[MAX_SHELLEY_ADDRESS_SIZE];
```
refi93, line 281: I had to restrict the address size to `MAX_SHELLEY_ADDRESS_SIZE` otherwise I was running into a stack overflow error, maybe you have additional insight @janmazak , but I have the suspicion that three concurrent hash contexts (tx, auxiliary data, catalyst payload) eat up just too much memory

refi93:
Upon further inspection I found the instruction state got prohibitively bloated with the addition of two new hash context which take 512 bytes together, decreasing the available stack size by 400 bytes which pushes the rest of logic to the edge of the available memory given that the overall available stack size is around 1500 bytes.

A solution to this would be moving the construction of the metadata before the rest of the sign transaction flow, creating a separate context for auxiliary data and the transaction body, so the overall state doesn't need to hold data about both (which would mean space for three hash contexts plus other shared info - just too much)

refi93:
closing in favor of https://github.com/vacuumlabs/ledger-app-cardano-shelley/pull/65

refi93:
closing in favor of #65

PR #64, opened 2021-04-07
Fix pool id UI

janmazak:


refi93 (approved):
LGTM, tested and works as expected

PR #65, opened 2021-04-08
Catalyst voting support

refi93:
Motivation: Add support for Catalyst voting registration

* based on top of global refactors from operator app PR which has been put on hold (this branch: https://github.com/vacuumlabs/ledger-app-cardano-shelley/tree/stackmemory), and Catalyst-specific changes come after commit https://github.com/vacuumlabs/ledger-app-cardano-shelley/pull/65/commits/a2ccb7deb9e28778ec8b2add3a25505e6a889c82

Changes:

* renamed metadata to auxData in the code to reflect latest CDDL
* introduce new hash builder for auxiliary data (former metadata in pre-Mary era) needed to compute the hash of the auxiliary data as such as well as the catalyst registration payload (needed to obtain the signature)
* introduce new dedicated submachine for Catalyst voting registration which triggers from the Auxiliary data step within the main signTx flow
* the catalyst registration submachine, once in its terminal state, returns the overall auxiliary data hash and registration signature in a single APDU
* split the sign transaction instruction context into three "parts" - `aux_data_ctx`, `body_ctx` and `witnesses_ctx`, to be able to reuse memory while serializing those three separate parts of the transaction, avoiding three concurrent hash contexts in case of the catalyst voting registration, otherwise Ledger was running out of memory
* APDU-wise, the auxiliary data APDU has been put right after the init APDU and extended with "auxiliary data type" enum, other than that there are no breaking changes (TBD if we want to renumber the handler codes https://github.com/vacuumlabs/ledger-app-cardano-shelley/pull/65/files#diff-5fb9b0dad635df88437117ed0aee787f834f3b8730b88ba5921f67b78f20fe52R1556)

Instruction state size:
We originally faced issues with a prohibitively big instruction state size (~1100 bytes) due to three independent hash contexts, each taking up 256 bytes. By putting processing of metadata before the rest of the transaction we were able to save space by sharing memory with the tx body hash builder, saving ~256 bytes.

The current state size is 816, previously it was 736, there are some opportunities for further optimization which should get us even closer to the original instruction state size:

* we can easily save 32 bytes by making the txHash and auxDataHash a union which I opted not to do now to not complicate the state further if we don't need to
* Another opportunity for saving 13 bytes is moving number of tx entries (inputs/outputs/withdrawals...) into the tx body part context (imho not worth it as it would completely change the init APDU, triggering a big diff and need for another APDU in the flow)
* save ~32 bytes by making the client send the staking path twice within the Catalyst APDU flow instead of persisting it in the context (first time to show/serialize the staking key, second time to obtain the signature) - IMHO not worth the sacrifice in signature-staking key consistency and we may optimize that relatively sensibly if we struggle with memory in the future

But if you see a good reason to do any of those optimizations now, feel free to raise it

Testing:

* added device test for auxiliary data hash builder
* ran yarn test-integration --grep signTxOrdinaryMary on vacuumlabs/ledgerjs-cardano-shelley#99 with a Catalyst testcase added


src/txHashBuilder.c
```diff
@@ -861,33 +861,33 @@ static void txHashBuilder_assertCanLeaveWithdrawals(tx_hash_builder_t* builder)
...
 
-void txHashBuilder_addMetadata(tx_hash_builder_t* builder, const uint8_t* metadataHashBuffer, size_t metadataHashBufferSize)
+void txHashBuilder_addAuxData(tx_hash_builder_t* builder, const uint8_t* auxDataHashBuffer, size_t auxDataHashBufferSize)
 {
 	txHashBuilder_assertCanLeaveWithdrawals(builder);
-	ASSERT(builder->includeMetadata);
+	ASSERT(builder->includeAuxData);
 
 	{
 		BUILDER_APPEND_CBOR(CBOR_TYPE_UNSIGNED, 7);
```
gabrielKerekes, line 870: The `TX_BODY_KEY_METADATA` constant should probably be used here. And it should also probably be renamed to `TX_BODY_KEY_AUXILIARY_DATA` then. Or maybe even `TX_BODY_KEY_AUXILIARY_DATA_HASH`.

gabrielKerekes (approved):
I've taken a look despite the fact, that it's probably going to be rewritten because of the memory problems.

I haven't found any problems, I just added a few curiosity comments.

src/signOpCert.h
```diff
@@ -0,0 +1,24 @@
+#ifndef H_CARDANO_APP_SIGN_OP_CERT
```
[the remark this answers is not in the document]
refi93: no, just a leftover from the rebase, will be deleted in the final version

src/addressUtilsShelley.c
```diff
@@ -503,3 +516,41 @@ void view_parseAddressParams(read_view_t* view, addressParams_t* params)
 		ASSERT(false);
 	}
 }
+
+static inline bool isSpendingPathConsistentWithAddressType(const address_type_t addressType, const bip44_path_t* spendingPath)
```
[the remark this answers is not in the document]
refi93: it most likely can, the code is currently a bit of a hybrid between the operator app branch where a few refactors regarding recognition of paths have been done and develop - we already agreed with @janmazak that he'll extract the refactors from the operator app to be usable here, eliminating this duplicity

src/uiScreens.c
```diff
@@ -23,6 +23,58 @@ void ui_displayPathScreen(
...
+
+	{
+		size_t len = strlen(out);
+		ASSERT(len + 1 < outSize);
+
+		bip44_printToStr(path, out + len, outSize - len);
+	}
+}
+
+void ui_displayAccountDescriptionScreen(
```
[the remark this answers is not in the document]
refi93: fixed - ui_displayAccountScreen has been removed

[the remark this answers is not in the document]
refi93: done - I made a function `_ui_displayPathAccountScreen` which takes a boolean flag whether to show the account description or nor and extracted the condition to `ui_displayStakingKeyScreen` and `ui_displayAccountScreen` outward-facing ui helpers

src/addressUtilsShelley.c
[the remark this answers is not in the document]
refi93: not sure if still relevant as I did a few rebases since then - now I see this function being used only within addressUtilsShelley.c - do you still see a point in reusing this function outside this file?

src/signTx.c
```diff
@@ -22,6 +22,40 @@ enum {
 };
 
 static ins_sign_tx_context_t* ctx = &(instructionState.signTxContext);
+static ins_sign_tx_body_context_t* txBodyCtx = &(instructionState.signTxContext.txPartCtx.body_ctx);
+static ins_sign_tx_aux_data_context_t* txAuxDataCtx = &(instructionState.signTxContext.txPartCtx.aux_data_ctx);
+static ins_sign_tx_witnesses_context_t* txWitnessesCtx = &(instructionState.signTxContext.txPartCtx.witnesses_ctx);
+
+// TODO - maybe add an enum to the global context which would specify the active tx part?
```
refi93, line 29: I'd appreciate your insight on this @janmazak

src/addressUtilsShelley.c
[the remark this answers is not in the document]
gabrielKerekes: Nope, it seems that the function has been removed from `securityPolicy.c` and is now only in `addressUtilsShelley.c`. :+1:

gabrielKerekes (approved):
Just one small thing. Otherwise LGTM.

src/cardano.h
```diff
@@ -15,25 +15,34 @@ STATIC_ASSERT(LOVELACE_MAX_SUPPLY < LOVELACE_INVALID, "bad LOVELACE_INVALID");
...
-#define METADATA_HASH_LENGTH 32
+#define AUX_DATA_HASH_LENGTH 32
+#define POOL_METADATA_HASH_LENGTH 32
+#define CATALYST_REGISTRATION_PAYLOAD_HASH_LENGTH 32
+#define ED25519_SIGNATURE_LENGTH 64
 
 #define MINTING_POLICY_ID_SIZE 28
 #define ASSET_NAME_SIZE_MAX 32
 
+#define REWARD_ACCOUNT_SIZE (1 + ADDRESS_KEY_HASH_LENGTH)
```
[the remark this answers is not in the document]
refi93: the reward "address" is not a proper address in the sense you could send funds there - I see that as the main rationale behind calling it an "account" instead. It's also how it's called in the Ledger specs: https://hydra.iohk.io/build/6026331/download/1/ledger-spec.pdf (page 9)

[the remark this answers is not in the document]
gabrielKerekes: [Here](https://docs.cardano.org/en/latest/learn/types-addresses.html#reward-account-addresses) they call it "Reward account address". There's mentions of "reward address" also in the ledger spec.

It's used in functions called `constructRewardAddressFromKeyPath` and even in that function it's then assigned to a constant called `ADDRESS_LENGTH`. It's used as the size for a variable called `rewardAddress`. `rewardAddress` is also used by withdrawals. It's also derived from addressParameters.

I think it would be less questionable to call it address, but I see that `rewardAccount` is also used for `poolParams` so I'm not sure it could be renamed at this point.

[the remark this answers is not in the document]
gabrielKerekes: In cardano-node they have `RewardAcnt` which gets serialised into `StakeAddress` :smile: Daedalus seems to be using reward account.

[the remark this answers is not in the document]
refi93: nevertheless, this is something this PR is not really touching, it's rather related to the refactors made in the operator app that have been extracted

src/auxDataHashBuilder.c
```diff
@@ -0,0 +1,250 @@
...
+	blake2b_256_append_cbor(&builder->auxDataHash, type, value, true); \
+	if (auxDataHashBuilder_isInCatalystRegistrationPayload(builder)) { \
+		blake2b_256_append_cbor(&builder->catalystRegistrationData.payloadHash, type, value, false); \
+	}
+
+#define BUILDER_APPEND_DATA(buffer, bufferSize) \
+	blake2b_256_append_and_trace(&builder->auxDataHash, buffer, bufferSize); \
+	if (auxDataHashBuilder_isInCatalystRegistrationPayload(builder)) { \
+		blake2b_256_append(&builder->catalystRegistrationData.payloadHash, buffer, bufferSize); \
+	}
```
janmazak, line 32: I find this quite hard to follow. There are some non-sugared calls to blake2b below, too (at least one). What about having macros like `APPEND_AUXDATA_CBOR`, `APPEND_AUXDATA_BUFFER`, `APPEND_CATALYST_CBOR`, `APPEND_BOTH_CBOR` (defined as composition of the two) etc.? Then at least one can use editor search function to see what is added to the respective hashes.

src/messageSigning.c
```diff
@@ -43,11 +43,9 @@ void getTxWitness(bip44_path_t* pathSpec,
 		TRY {
 			derivePrivateKey(pathSpec, &chainCode, &privateKey);
```
janmazak, line 44: Imagine having a `noinline` function wrapping `derivePrivateKey` (which would more accurately be called `deriveExtendedPrivateKey`) and throwing away the chain code which we never need for signing. I think it will save some stack space. Might be handy in the future.

refi93: IMHO this is enough to do when we face actual stack space shortage while signing and do a proper analysis of the root cause should that be the case. I'd rather not trade-off complexity for memory usage if we don't urgently need it

src/auxDataHashBuilder_test.c
```diff
@@ -0,0 +1,96 @@
...
+		uint8_t tmp[32];
+		size_t tmpSize = decode_hex(stakingKey, tmp, SIZEOF(tmp));
+		auxDataHashBuilder_catalystRegistration_addStakingKey(
+		        &builder,
+		        tmp, tmpSize
+		);
+	}
+
+	{
+		uint8_t tmp[86];
```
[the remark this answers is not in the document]
refi93: the voting rewards address is a standard base address, so it's 57 I took the number 86 from tx hash builder tests if I'm not mistaken, but 57 should be enough and makes sense to make it as tight as possible to crash if the output happens to be bigger than expected

src/auxDataHashBuilder.c
refi93: This could work but I feel like it would be kind of cumbersome to manage/extend as soon as we'd want to add anything else, ultimately creating a tech debt for whoever would add support for some other kind of metadata which is plausible to happen in the foreseeable future.

I ended up creating a binary mask backed by an enum, specifying what hash builders to select which allows to explicitly select the relevant hash builders by their codename when using the macro - I just pushed it. What do you think?

src/auxDataHashBuilder.c
```diff
@@ -164,10 +148,10 @@ void auxDataHashBuilder_catalystRegistration_addVotingRewardsAddress(
 
 	ASSERT(builder->state == AUX_DATA_HASH_BUILDER_IN_CATALYST_REGISTRATION_PAYLOAD_STAKING_KEY);
```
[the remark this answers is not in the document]
refi93: fair point - until there's nothing else related to Catalyst, it should be self-evident even without a comment. I shortened it to catalyst_payload and catalyst_signature

src/signTx.c
refi93: as discussed offline, controlling access to the corresponding tx part based on an enum and wrapping accessor function seems worthwile, but it would bloat this PR with too many changes, so we'd do it in another PR - I made an issue for it: https://github.com/vacuumlabs/ledger-app-cardano-shelley/issues/66

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+			getCatalystVotingRegistrationSignature(
+			        &subctx->stakingKeyPath,
+			        votingPayloadHashBuffer, CATALYST_REGISTRATION_PAYLOAD_HASH_LENGTH,
+			        (uint8_t*) &subctx->stateData.registrationSignature, ED25519_SIGNATURE_LENGTH
+			);
+		}
+		auxDataHashBuilder_catalystRegistration_addSignature(auxDataHashBuilder, (uint8_t*) &subctx->stateData.registrationSignature, ED25519_SIGNATURE_LENGTH);
+		auxDataHashBuilder_catalystRegistration_addAuxiliaryScripts(auxDataHashBuilder);
+
+		auxDataHashBuilder_finalize(auxDataHashBuilder, (uint8_t*) &ctx->auxDataHash, AUX_DATA_HASH_LENGTH);
```
janmazak, line 515: We do the `(uint8_t*)` quite often here and it looks suspicious. AFAIK we should use
```ctx->auxDataHash```
instead of 
```(uint8_t*) &ctx->auxDataHash```, probably everywhere.

Comment (janmazak) [reply src/signTxCatalystRegistration.c:515]: Unrelated to this, github does not show the right place at this point to me, but there is a compiler warning:
```
src/auxDataHashBuilder.c:20:1: warning: 'static' ignored on this declaration [-Wmissing-declarations]
static enum {
```

Comment (janmazak) [reply src/signTxCatalystRegistration.c:515]: Another unrelated:
```
	txHashBuilder_assertCanLeaveAuxData(builder);

	// add fee item into the main tx body map
```
the comment should not be about fee.

File: src/auxDataHashBuilder.c [hunk 12]
```diff
@@ -18,18 +18,18 @@
 #endif // DEVEL
 
 static enum {
-	HASH_CTX_AUX_DATA = (1u << 0),
-	HASH_CTX_CATALYST_PAYLOAD = (1u << 1)
+	HC_AUX_DATA = (1u << 0),
+	HC_CATALYST_PAYLOAD = (1u << 1)
```
janmazak:
```suggestion
	HC_AUX_DATA =         (1u << 0),
	HC_CATALYST_PAYLOAD = (1u << 1)
```

src/signTx.c
```diff
@@ -432,6 +488,150 @@ static void signTx_handleInitAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
...
+			break;
+		default:
+			THROW(ERR_INVALID_DATA);
+		}
+
+		ASSERT(view_remainingSize(&view) == 0);
+	}
+
+
+	security_policy_t policy = policyForSignTxAuxData();
```
janmazak, line 599: I would consider adding auxDataType as an argument here, even if we do not use it. That way, we explicitly signalize our intent, otherwise it looks a bit like we might have forgot that there are two cases to distinguish.

src/signTx.c
```diff
@@ -1343,37 +1468,41 @@ static void signTx_handleWitness_ui_runStep()
...
 		TRACE("Sending witness data");
-		TRACE_BUFFER(ctx->stageData.witness.signature, SIZEOF(ctx->stageData.witness.signature));
-		io_send_buf(SUCCESS, ctx->stageData.witness.signature, SIZEOF(ctx->stageData.witness.signature));
+		TRACE_BUFFER(txWitnessesCtx->stageData.witness.signature, SIZEOF(txWitnessesCtx->stageData.witness.signature));
+		io_send_buf(SUCCESS, txWitnessesCtx->stageData.witness.signature, SIZEOF(txWitnessesCtx->stageData.witness.signature));
 		ui_displayBusy(); // needs to happen after I/O
 
-		ctx->currentWitness++;
-		if (ctx->currentWitness == ctx->numWitnesses) {
+		TRACE("CUR WIT %d vs NUM WIT %d", txWitnessesCtx->currentWitness, ctx->numWitnesses);
```
janmazak, line 1475: Is this just for temporary debugging?

src/signTx.c
```diff
@@ -1427,13 +1556,14 @@ static subhandler_fn_t* lookup_subhandler(uint8_t p1)
 #	define  CASE(P1, HANDLER) case P1: return HANDLER;
 #	define  DEFAULT(HANDLER)  default: return HANDLER;
 		CASE(0x01, signTx_handleInitAPDU);
+		// APDU flow out of original order to optimize memory usage related to serializing auxiliary data
```
janmazak, line 1559: We shouldn't refer to "original order". What about
```
// Auxiliary data have to be handled before tx body because of memory consumption:
// in certain cases we need compute a rolling hash, 
// and that cannot be done while the computation of tx body hash is in progress.
```

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+		VALIDATE(wireDataSize == SIZEOF(subctx->stateData.votingPubKey), ERR_INVALID_DATA);
+
+		{
+			STATIC_ASSERT(SIZEOF(subctx->stateData.votingPubKey) == CATALYST_VOTING_PUBLIC_KEY_LENGTH, "wrong voting public key size");
+			os_memmove(subctx->stateData.votingPubKey, wireDataBuffer, CATALYST_VOTING_PUBLIC_KEY_LENGTH);
+		}
+	}
+	{
+		auxDataHashBuilder_catalystRegistration_enter(auxDataHashBuilder);
+		auxDataHashBuilder_catalystRegistration_enterPayload(auxDataHashBuilder);
```
janmazak, line 148: I would consider moving such things into `advanceState`. That's how we do it in signTx.c (handle*APDU only passes currently received data to txhashbuilder).

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+	UI_STEP_BEGIN(subctx->ui_step);
+
+	UI_STEP(HANDLE_STAKING_KEY_STEP_WARNING) {
+		ui_displayPaginatedText(
+		        "Unusual request",
+		        "Proceed with care",
+		        this_fn
+		);
+	}
+
```
janmazak, line 183: empty line

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+		explicit_bzero(&subctx->stateData, SIZEOF(subctx->stateData));
+	}
+	{
+		TRACE_BUFFER(wireDataBuffer, wireDataSize);
+
+		VALIDATE(wireDataSize == SIZEOF(subctx->stateData.votingPubKey), ERR_INVALID_DATA);
+
+		{
+			STATIC_ASSERT(SIZEOF(subctx->stateData.votingPubKey) == CATALYST_VOTING_PUBLIC_KEY_LENGTH, "wrong voting public key size");
+			os_memmove(subctx->stateData.votingPubKey, wireDataBuffer, CATALYST_VOTING_PUBLIC_KEY_LENGTH);
```
janmazak, line 143: Using `view_memmove` is preferable IMHO. Might be worth creating an issue and fixing it consistently across all the codebase in a separate PR.

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+		ui_displayPaginatedText(
+		        "Unusual request",
+		        "Proceed with care",
+		        this_fn
+		);
+	}
+	UI_STEP(HANDLE_VOTING_REWARDS_ADDRESS_PARAMS_STEP_DISPLAY_ADDRESS) {
+		uint8_t addressBuffer[MAX_ADDRESS_SIZE];
+		size_t addressSize;
+		addressSize = deriveAddress(
```
janmazak, line 282: Maybe merge these into one (initialize `addressSize` straight away).

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+static void signTxCatalystRegistration_handleNonce_ui_runStep()
+{
+	TRACE("UI step %d", subctx->ui_step);
+	TRACE_STACK_USAGE();
+	ui_callback_fn_t* this_fn = signTxCatalystRegistration_handleNonce_ui_runStep;
+
+	UI_STEP_BEGIN(subctx->ui_step);
+
+	UI_STEP(HANDLE_NONCE_STEP_DISPLAY) {
+		STATIC_ASSERT(SIZEOF(subctx->stateData.votingPubKey) == CATALYST_VOTING_PUBLIC_KEY_LENGTH, "wrong voting public key size");
```
janmazak, line 387: leftover from copy-paste

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+
+// ============================== NONCE ==============================
+
+enum {
+	HANDLE_NONCE_STEP_DISPLAY = 8600,
+	HANDLE_NONCE_STEP_RESPOND,
+	HANDLE_NONCE_STEP_INVALID,
+};
+
+__noinline_due_to_stack__
```
janmazak, line 377: unnecessary, does not allocate buffers inside

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+		TRACE(
+		        "Catalyst registration nonce: %d",
+		        subctx->stateData.nonce
+		);
+	}
+	{
+		auxDataHashBuilder_catalystRegistration_addNonce(auxDataHashBuilder, subctx->stateData.nonce);
+	}
+
+	// TODO - is it worth declaring a policy for this?
```
janmazak, line 428: I think it is an advantage to have all security policies in one place. It is definitely easier to deal with a change in security policy if it has an explicit name. Then it is much easier to find all the relevant code (compared to a plain variable assignment) even if meaning of the policy is not entirely clear from its name.

src/signTx.c
refi93: yes, sorry I forgot it here, removing

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+		        this_fn
+		);
+	}
+	UI_STEP(HANDLE_CONFIRM_STEP_RESPOND) {
+		struct {
+			uint8_t auxDataHash[AUX_DATA_HASH_LENGTH];
+			uint8_t signature[ED25519_SIGNATURE_LENGTH];
+		} wireResponse;
+
+		ASSERT(SIZEOF(ctx->auxDataHash) == AUX_DATA_HASH_LENGTH);
```
janmazak, line 471: What about asserting `SIZEOF(ctx->auxDataHash)` too? Also couple of lines below. And the asserts can be made static. 
(Might be worth checking if they are static across all the codebase.)

janmazak (commented):
That's all for now.

src/signTx.c
```diff
@@ -432,6 +488,150 @@ static void signTx_handleInitAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
...
+
+		switch (txAuxDataCtx->auxDataType) {
+		case AUX_DATA_TYPE_ARBITRARY_HASH:
+			// parse data
+			VALIDATE(view_remainingSize(&view) == AUX_DATA_HASH_LENGTH, ERR_INVALID_DATA);
+			STATIC_ASSERT(SIZEOF(ctx->auxDataHash) == AUX_DATA_HASH_LENGTH, "wrong metadata hash length");
+			view_memmove(ctx->auxDataHash, &view, AUX_DATA_HASH_LENGTH);
+			txAuxDataCtx->auxDataReceived = true;
+			break;
+		case AUX_DATA_TYPE_CATALYST_VOTING_KEY_REGISTRATION:
```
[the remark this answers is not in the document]
refi93: this is being validated below the switch statement, to cover both the arbitrary hash and catalyst registration case

src/signTxCatalystRegistration.c
refi93: This line above is doing just that - did you have something else in mind? I changed it to a static assert nevertheless

src/signTx.h
```diff
@@ -50,69 +52,92 @@ typedef struct {
...
+	uint16_t numCertificates;
+	uint16_t numWithdrawals; // reward withdrawals
+	bool includeValidityIntervalStart;
+	uint16_t numWitnesses;
+	bool includeAuxData;
+
+	uint8_t auxDataHash[TX_HASH_LENGTH];
+	uint8_t txHash[TX_HASH_LENGTH];
+
+	common_tx_data_t commonTxData;
```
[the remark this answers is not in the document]
refi93: agree - I'd leave that for operator app fixups as Catalyst isn't really dealing with this property

src/signTx.h
```diff
@@ -50,69 +52,92 @@ typedef struct {
...
 		output_context_t output_subctx;
 	} stageContext;
+} ins_sign_tx_body_context_t;
+
+typedef struct {
+	uint16_t currentWitness;
+	struct {
+		sign_tx_witness_data_t witness;
+	} stageData;
+} ins_sign_tx_witnesses_context_t;
```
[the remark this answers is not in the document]
refi93: I ordered it as it is done in the CDDL but makes sense to preserve the APDU order

src/signTxCatalystRegistration.c
refi93: The advanceState function is called after the first APDU that already handled the voting key and is supposed to add it to the hash context. It could be added to advanceState function in the metadata step in signTx but this would IMHO leak responsibilities from the submachine. Another solution could be making that part of the initialization of the aux data hash builder, but that also doesn't seem OK as the aux data hash builder is meant potentially for any metadata, not just a catalyst registration, so this is the most appropriate place I found after all.

Do you see how to introduce those "enter steps" sensibly into advanceState?

src/signTxCatalystRegistration.c
refi93: done,  I created basic security policies for all Catalyst registration steps

src/signTxCatalystRegistration.c
```diff
@@ -0,0 +1,576 @@
...
+#include "textUtils.h"
+#include "bufView.h"
+#include "securityPolicy.h"
+#include "messageSigning.h"
+
+// we want to distinguish the two state machines to avoid potential confusion:
+// ctx / subctx
+// stage / state
+// from ctx, we only make the necessary parts available to avoid mistaken overwrites
+static ins_sign_tx_context_t* ctx = &(instructionState.signTxContext);
```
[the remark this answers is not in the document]
refi93: It's actually used in two functions in multiple places albeit for the same thing - dealing with the auxDataHash stored in the top-level instruction state in order to propagate it to signTx function which is responsible of including it in the tx body in a later step. So I'm not sure if it would be practical to hide it within the functions where it's being used.

But I agree it makes sense to do the initialization of the subctx the way it's done in the operator app, in order to prevent mixups of contexts

src/auxData.h
```diff
@@ -0,0 +1,9 @@
+#ifndef H_CARDANO_APP_AUX_DATA
+#define H_CARDANO_APP_AUX_DATA
+
+typedef enum {
+	AUX_DATA_TYPE_ARBITRARY_HASH = 0,
+	AUX_DATA_TYPE_CATALYST_VOTING_KEY_REGISTRATION = 1,
+} aux_data_type_t;
+
+#endif // H_CARDANO_APP_AUX_DATA
```
janmazak, line 9: I'd prefer to name the file `signTxAuxData.h`.

src/signTx.c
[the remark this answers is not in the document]
janmazak: There is an ASSERT, not VALIDATE.

src/signTxCatalystRegistration.c
janmazak: I guess I meant the other variable (source/destination of os_memmove), it it not relevat anymore with view_memmove.

janmazak (approved):

PR #70, opened 2021-04-26
Catalyst: remove warning for reward address

janmazak:


refi93 (approved):
Tested with ledgerjs integration tests and works as expected - warning for rewards address is not present

PR #72, opened 2021-05-13
introduce size and lexicographical based ordering on multiasset scrip…

mkv-vcm:
…t hashes and token ids

obviously without the debug data. Wrote my questions into the code as TODOs.


mkv-vcm:
@gabrielKerekes @janmazak please have a look, especially the TODO'd parts.

src/signTxOutput.c
```diff
@@ -367,6 +367,27 @@ static void signTxOutput_handleAssetGroup_ui_runStep()
 	UI_STEP_END(HANDLE_ASSET_GROUP_STEP_INVALID);
 }
 
+static int signTxOutput_lexiCompareViews(read_view_t lhs, read_view_t rhs)
```
gabrielKerekes, line 370: I think this could perhaps be moved to `signTxUtils.c` (or somewhere else) as it is a pretty generic function.

gabrielKerekes (commented):
I wasn't able to get it to fail by passing in the same policy id twice e.g.:
```Javascript
tokenBundle: [
  {
    policyIdHex: "95a292ffee938be03e9bae5657982a74e9014eb4960108c9e23a5b39",
    tokens: [
      {
        assetNameHex: "64652474436f696e",
        amount: "7878754",
      },
      {     
        assetNameHex: "74652474436f696e",
        amount: "7878754",
      }
    ],
  },
  {
    policyIdHex: "95a292ffee938be03e9bae5657982a74e9014eb4960108c9e23a5b39",
    tokens: [
      {
        assetNameHex: "",
        amount: "78787",
      },
      {
        assetNameHex: "00",
        amount: "78787",
      }
    ],
  },
]
```
This should fail because of the policies being the same, no? It also doesn't fail if the first policy is changed to `ffa292ffee938be03e9bae5657982a74e9014eb4960108c9e23a5b39`. But it should fail, shouldn't it? Am I missing something?

It does work as expected with asset names though.

src/signTxOutput.c
mkv-vcm: It is generic, but I'm not sure where else it would be used, and I thought to keep it as local as possible until the need arises. But I would be interested in Mazák's opinion on this.

mkv-vcm:
> I wasn't able to get it to fail by passing in the same policy id twice e.g.:

Not sure what made me not notice this, but yeah. It's resolved now.

src/signTxOutput.c
```diff
@@ -384,6 +405,13 @@ static void signTxOutput_handleAssetGroupAPDU(uint8_t* wireDataBuffer, size_t wi
 
 		VALIDATE(view_remainingSize(&view) >= MINTING_POLICY_ID_SIZE, ERR_INVALID_DATA);
 		STATIC_ASSERT(SIZEOF(tokenGroup->policyId) == MINTING_POLICY_ID_SIZE, "wrong policy id size");
+		PRINTF("scrp 1 >%.*H<\n", MINTING_POLICY_ID_SIZE, wireDataBuffer);
+		PRINTF("scrp 2 >%.*H<\n", MINTING_POLICY_ID_SIZE, tokenGroup->policyId);
+		VALIDATE(signTxOutput_lexiCompareViews(
+			make_read_view(tokenGroup->policyId, tokenGroup->policyId + MINTING_POLICY_ID_SIZE),
+			make_read_view(wireDataBuffer, wireDataBuffer + MINTING_POLICY_ID_SIZE)
```
janmazak, line 412: I don't like working with `wireDataBuffer` directly, it duplicates the parsing logic (if APDU data format is changed, we will need to modify two different places) and it is not immediately obvious what we do here. Let's consider an alternative:
creating a local buffer where we copy the policyId, then validate it (i.e. compare with previous), and then memcpy it to its proper place. What do you think? (maybe a one-line comment explaining that policyIds have to be ordered properly in the map in tx body)

src/signTxOutput.c
janmazak: What is the advantage of using read_views here instead of passing buffer + its size directly? (You can pass a wrong buffer size, but it's the same with the view, you can create a wrong one.)

janmazak: AFAIK we will need this for withdrawals and token minting, too, and in principle, it is a comparison between two cbor map keys, so I would consider putting it in `cbor.h` and naming it accordingly (perhaps not as a general comparison, but just a function returning bool iff the keys are in the correct order).

src/signTxOutput.c
```diff
@@ -404,6 +432,7 @@ static void signTxOutput_handleAssetGroupAPDU(uint8_t* wireDataBuffer, size_t wi
 		);
 		TRACE();
 	}
+	subctx->stateData.token.assetNameSize = ASSET_NAME_SIZE_MAX + 1;	//maybe turn it into a signed number and use -1? takes up too much space anyways
```
janmazak, line 435: I am sure we do not want to use raw int for buffer size in just a single place.

I don't understand this assignment: why is it in this place? And I really dislike using buffer size that exceeds the max. Can we use `subctx->currentToken` to distinguish whether we are at the first policyId or not?

src/signTxOutput.c
```diff
@@ -468,10 +497,20 @@ static void signTxOutput_handleTokenAPDU(uint8_t* wireDataBuffer, size_t wireDat
...
 
 		VALIDATE(view_remainingSize(&view) >= 4, ERR_INVALID_DATA);
-		token->assetNameSize = parse_u4be(&view);
+		const size_t newAssetNameSize = parse_u4be(&view);
+		PRINTF("  asst 1 >%.*H<\n", wireDataSize, wireDataBuffer);
+		PRINTF("  asst 2 >%d<\n", token->assetNameSize);
+		VALIDATE(ASSET_NAME_SIZE_MAX + 1 == token->assetNameSize || signTxOutput_lexiCompareViews(
+			make_read_view(token->assetNameBytes, token->assetNameBytes + token->assetNameSize),
+			make_read_view(wireDataBuffer + 4, wireDataBuffer + 4 + newAssetNameSize)
+			) < 0, ERR_INVALID_DATA);	//TODO should be STATE?
```
janmazak, line 506: `ERR_INVALID_STATE` is used for incorrect APDU, not just invalid data content of the APDU.

Perhaps you can replace the comment
`	// Request is not valid in the context of previous calls`
with
`	// Request is not valid in the context of previous calls (typically, wrong P1 or P2 in APDU msg)`

src/signTxOutput.c
```diff
@@ -468,10 +497,20 @@ static void signTxOutput_handleTokenAPDU(uint8_t* wireDataBuffer, size_t wireDat
...
+		PRINTF("  asst 3\n");
```
janmazak, line 507: I try to avoid cryptic names (asst vs. asset).

src/signTxOutput.c
mkv-vcm: True, that reduces the code duplication, but increases the stack size.

src/signTxOutput.c
mkv-vcm: > What is the advantage of using read_views here instead of passing buffer + its size directly? (You can pass a wrong buffer size, but it's the same with the view, you can create a wrong one.)

What is the advantage anywhere?

src/signTxOutput.c
mkv-vcm: I don't plan to leave any of those prints in the final version.

src/signTxOutput.c
janmazak: For read views, it's mostly easier parsing of bip44 paths because those have variable length (same with address params). Especially when several of those are coupled together (e.g. wireBuffer containing two paths).

src/signTxOutput.c
janmazak: It's possible to wrap it in a noinline_due_to_stack function, say, `_parseTokenGroupPolicyId`.

src/signTxOutput.c
```diff
@@ -468,11 +472,20 @@ static void signTxOutput_handleTokenAPDU(uint8_t* wireDataBuffer, size_t wireDat
...
+
+		view_memmove(candidateAssetNameBytes, &view, candidateAssetNameSize);
+		VALIDATE(0 == subctx->currentToken || cbor_mapKeyFulfillsCanonicalOrdering(
+			token->assetNameBytes, token->assetNameSize,
+			candidateAssetNameBytes, candidateAssetNameSize) < 0, ERR_INVALID_DATA);
 
-		ASSERT(token->assetNameSize <= SIZEOF(token->assetNameBytes));
-		view_memmove(token->assetNameBytes, &view, token->assetNameSize);
+		token->assetNameSize = candidateAssetNameSize;
+		memcpy(token->assetNameBytes, candidateAssetNameBytes, candidateAssetNameSize);
```
janmazak, line 488: Before `memcpy/memmove`, always add an assertion that there is enough space in the target buffer.
(Several times in this PR.)

src/cbor.c
```diff
@@ -154,3 +154,21 @@ size_t cbor_writeToken(uint8_t type, uint64_t value, uint8_t* buffer, size_t buf
 	}
 #undef CHECK_BUF_LEN
 }
+
+bool cbor_mapKeyFulfillsCanonicalOrdering(const uint8_t* previousBuffer, size_t previousSize,
+										const uint8_t* nextBuffer, size_t nextSize)
+{
+	if(previousSize != nextSize) {
```
janmazak, line 161: We use space after if, for etc.

janmazak (commented):
The same ordering should be checked for withdrawals, however, that would cost memory (in tx_hash_builder_t). So I'd suggest that you wrote a comment in the function `txHashBuilder_addWithdrawal` where we explain that we do not call `cbor_mapKeyFulfillsCanonicalOrdering` (albeit it would be reasonable) because we don't want to increase memory footprint of `tx_hash_builder_t`.

src/signTxOutput.c
```diff
@@ -468,11 +472,20 @@ static void signTxOutput_handleTokenAPDU(uint8_t* wireDataBuffer, size_t wireDat
 		read_view_t view = make_read_view(wireDataBuffer, wireDataBuffer + wireDataSize);
 
 		VALIDATE(view_remainingSize(&view) >= 4, ERR_INVALID_DATA);
-		token->assetNameSize = parse_u4be(&view);
-		VALIDATE(token->assetNameSize <= ASSET_NAME_SIZE_MAX, ERR_INVALID_DATA);
+		const size_t candidateAssetNameSize = parse_u4be(&view);
+
+		VALIDATE(candidateAssetNameSize <= ASSET_NAME_SIZE_MAX, ERR_INVALID_DATA);
+		ASSERT(candidateAssetNameSize <= SIZEOF(token->assetNameBytes));
+		//TODO not sure if the C99 dynamic array should be used on this project?
```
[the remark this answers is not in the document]
mkv-vcm: I remembered that variable length arrays were a C99 feature, but it seems they were pushed to C11 and even there they are not universally supported.
https://en.wikipedia.org/wiki/Variable-length_array

Not that it matters much, we have to be able to support the max length anyways.

src/cbor.h
```diff
@@ -55,6 +55,8 @@ size_t cbor_writeToken(uint8_t type, uint64_t value, uint8_t* buffer, size_t buf
 
 cbor_token_t cbor_parseToken(const uint8_t* buf, size_t size);
 
+bool cbor_mapKeyFulfillsCanonicalOrdering(const uint8_t* previousBuffer, size_t previousSize,
```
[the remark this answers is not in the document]
mkv-vcm: Even view_memmove?

src/signTxOutput.c
mkv-vcm: Even view_memmove?

src/signTxOutput.c
```diff
@@ -384,7 +384,14 @@ static void signTxOutput_handleAssetGroupAPDU(uint8_t* wireDataBuffer, size_t wi
...
 		VALIDATE(view_remainingSize(&view) >= MINTING_POLICY_ID_SIZE, ERR_INVALID_DATA);
 		STATIC_ASSERT(SIZEOF(tokenGroup->policyId) == MINTING_POLICY_ID_SIZE, "wrong policy id size");
-		view_memmove(tokenGroup->policyId, &view, MINTING_POLICY_ID_SIZE);
+		uint8_t candidatePolicyId[MINTING_POLICY_ID_SIZE];
+		view_memmove(candidatePolicyId, &view, MINTING_POLICY_ID_SIZE);
+		VALIDATE(0 == subctx->currentAssetGroup || cbor_mapKeyFulfillsCanonicalOrdering(
+		                 tokenGroup->policyId, MINTING_POLICY_ID_SIZE,
+		                 candidatePolicyId, MINTING_POLICY_ID_SIZE), ERR_INVALID_DATA);
+
+		ASSERT(sizeof(tokenGroup->policyId) >= MINTING_POLICY_ID_SIZE);
```
janmazak, line 393: We can use static (compile-time) asserts and `SIZEOF` instead of `sizeof` where it is possible, for example:
```
			STATIC_ASSERT(SIZEOF(ctx->auxDataHash) == AUX_DATA_HASH_LENGTH, "wrong auxiliary data hash length");
			view_memmove(ctx->auxDataHash, &view, AUX_DATA_HASH_LENGTH);
```

src/txHashBuilder.c
```diff
@@ -821,6 +821,8 @@ void txHashBuilder_addWithdrawal(
         uint64_t amount
 )
 {
+	//should call cbor_mapKeyFulfillsCanonicalOrdering for withdrawals,
+	//	but the memory footprint would be too much
```
janmazak, line 825:
```suggestion
	// should call cbor_mapKeyFulfillsCanonicalOrdering for withdrawals,
	// but storing the previous withdrawal would increase the memory footprint of tx_hash_builder_t too much
```

janmazak (approved):
Please squash it into a single commit and rebase it on top of `operatorapp` (we want to make a common PR with both changes; it should go without any conflicts I think).

gabrielKerekes:
I tried running some custom tests and tx signing with a single valid `tokenBundle` seems to fail in FW. I might be doing something wrong though, but multiasset tests also seem to fail for me. Could you just verify please that the tests pass and that you can sign a transaction with the following token bundle, which IMHO should be valid? It passes through LedgerJS (with the ordering already enforced) but fails on Ledger with `Invalid data supplied to Ledger`. 
```
tokenBundle: [
  {
    policyIdHex:
      "95a292ffee938be03e9bae5657982a74e9014eb4960108c9e23a5b37",
    tokens: [
      {
        assetNameHex: "74652474436f69",
        amount: "7878754",
      },
      {
        assetNameHex: "74652474436f696e",
        amount: "7878754",
      },
    ],
  },
],
```
My tests output:
```
  signTxOrdinaryMary
    1) Mary era transaction with a multiasset output
    2) Mary era transaction with a complex multiasset output
    ✓ Mary era transaction with big numbers (14595ms)
```

mkv-vcm:
> I tried running some custom tests and tx signing with a single valid `tokenBundle` seems to fail in FW. I might be doing something wrong though, but multiasset tests also seem to fail for me. Could you just verify please that the tests pass and that you can sign a transaction with the following token bundle, which IMHO should be valid? It passes through LedgerJS (with the ordering already enforced) but fails on Ledger with `Invalid data supplied to Ledger`. The same `tokenBundle` works OK on operator app FW.

You are correct, it was a bad implementation. After I switched from compare (where 0 is the equal) to a bool return function, I forgot to take out the < 0 after the function call for assets. Now it's fixed.

gabrielKerekes (approved):

janmazak:
Will be merged together with operator app.

PR #74, opened 2021-05-18
implement int64 string formatting

mkv-vcm:
String form of int64 type.

### DIFF
--- a/src/textUtils.c
+++ b/src/textUtils.c
@@ -59,40 +59,79 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
 	return rawSize + suffixLength;
 }
 
+static size_t stringifyUint64ToBufferReverse(uint64_t number, char* buffer, size_t bufferSize)
+{
+	char* currChar = buffer;
+	char* const end = buffer + bufferSize;
+
+	// We print in reverse
+	// We want at least one iteration
+	size_t printedChars = 0;
+	do {
+		WRITE_CHAR(currChar, end, '0' + (number % 10));
+		number /= 10;
+		++printedChars;
+	} while (number > 0);
+	WRITE_CHAR(currChar, end, '\0');
+	return printedChars;
+}
+
+static void printReversedStringToBuffer(const char* reversed, char* out, size_t outSize)
+{
+	const size_t reversedSize = strlen(reversed);
+	if (reversedSize + 1 > outSize) {
+		THROW(ERR_DATA_TOO_LARGE);
+	}
+	for (size_t i = 0; i < reversedSize; i++) {
+		out[i] = reversed[reversedSize - 1 - i];
+	}
+	out[reversedSize] = 0;
+
+	ASSERT(strlen(out) == reversedSize);
+}
+
 size_t str_formatUint64(uint64_t number, char* out, size_t outSize)
 {
 	ASSERT(outSize < BUFFER_SIZE_PARANOIA);
 
-	char scratchBuffer[30];
-	char* ptr = BEGIN(scratchBuffer);
-	char* end = END(scratchBuffer);
+	{
+		char tmpReversed[30];
 
-	// We print in reverse
-	// We want at least one iteration
-	int place = 0;
-	do {
-		WRITE_CHAR(ptr, end, '0' + (number % 10));
-		number /= 10;
-		place++;
-	} while (number > 0);
-
-	// Size without terminating character
-	STATIC_ASSERT(sizeof(ptr - scratchBuffer) == sizeof(size_t), "bad size_t size");
-	size_t rawSize = (size_t) (ptr - scratchBuffer);
-
-	if (rawSize + 1 > outSize) {
-		THROW(ERR_DATA_TOO_LARGE);
+		stringifyUint64ToBufferReverse(number, tmpReversed, SIZEOF(tmpReversed));
+		const size_t reversedSize = strlen(tmpReversed);
+		printReversedStringToBuffer(tmpReversed, out, outSize);
+		ASSERT(strlen(out) == reversedSize);
 	}
+	return strlen(out);
+}
 
-	// Copy reversed & append terminator
-	for (size_t i = 0; i < rawSize; i++) {
-		out[i] = scratchBuffer[rawSize - 1 - i];
+size_t str_formatInt64(int64_t number, char* out, size_t outSize)
+{
+	ASSERT(outSize < BUFFER_SIZE_PARANOIA);
+
+	uint64_t signlessNumber;
+	if (number < 0) {
+		if (INT64_MIN == number) {
+			signlessNumber = ((uint64_t)INT64_MAX) + 1;
+		} else {
+			signlessNumber = (uint64_t)(-number);
+		}
+	} else {
+		signlessNumber = (uint64_t)number;
 	}
-	out[rawSize] = 0;
-
-	ASSERT(strlen(out) == rawSize);
-
-	return rawSize;
+	{
+		char tmpReversed[30];
+		//size without the potential '-' sign
+		stringifyUint64ToBufferReverse(signlessNumber, tmpReversed, SIZEOF(tmpReversed) - 1);
+		size_t reversedLength = strlen(tmpReversed);
+		if (number < 0) {
+			snprintf(tmpReversed + reversedLength, SIZEOF(tmpReversed) - reversedLength, "%c", '-');
+			++reversedLength;
+		}
+		printReversedStringToBuffer(tmpReversed, out, outSize);
+		ASSERT(strlen(out) == reversedLength);
+	}
+	return strlen(out);
 }
 
 #ifdef DEVEL
--- a/src/cbor.c
+++ b/src/cbor.c
@@ -154,3 +154,23 @@ size_t cbor_writeToken(uint8_t type, uint64_t value, uint8_t* buffer, size_t buf
 	}
 #undef CHECK_BUF_LEN
 }
+
+bool cbor_mapKeyFulfillsCanonicalOrdering(
+        const uint8_t* previousBuffer, size_t previousSize,
+        const uint8_t* nextBuffer, size_t nextSize
+)
+{
+	if (previousSize != nextSize) {
+		return previousSize < nextSize;
+	}
+	for (size_t i = 0; i < previousSize; ++i) {
+		if (*previousBuffer != *nextBuffer) {
+			return *previousBuffer < *nextBuffer;
+		}
+		++previousBuffer;
+		++nextBuffer;
+	}
+	// key duplication is an error
+	return false;
+}
+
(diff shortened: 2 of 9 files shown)


src/textUtils.c
```diff
@@ -59,39 +59,70 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
...
+}
 
+size_t str_formatInt64(int64_t number, char* out, size_t outSize)
+{
+	ASSERT(outSize < BUFFER_SIZE_PARANOIA);
+
+	char scratchBuffer[30];
+	char* ptr = BEGIN(scratchBuffer);
+	char* end = END(scratchBuffer);
+
```
janmazak, line 109: I'd only define these where we actually need them, and it is generally safer to limit scopes of variables (as you can see in many other places, the codebase is very liberal with {} brackets).

src/textUtils.c
```diff
@@ -59,39 +59,70 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
...
+		} else {
+			signlessNumber = (uint64_t) llabs(number);
+		}
+	} else {
+		signlessNumber = (uint64_t) number;
+	}
+	// Size without terminating character (and potential sign)
+	size_t rawSize = stringifyUint64ToBufferReverse(signlessNumber, ptr, end - 1);
+	if (number < 0) {
+		scratchBuffer[rawSize++] = '-';
```
janmazak, line 123: Missing `'\0'` at the end of the string? 
I think it is safer to use `snprintf` (take a look at `str_formatAdaAmount` or `bip44_printToStr` for an example).

janmazak: And before calling `printReversedStringToBuffer`, we can assert that rawSize is what it should be, i.e. `strlen(scratchBuffer)`.

src/textUtils.c
```diff
@@ -59,39 +59,70 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
...
+	const size_t rawSize = stringifyUint64ToBufferReverse(number, ptr, end);
+	printReversedStringToBuffer(scratchBuffer, rawSize, out, outSize);
+	return rawSize;
+}
 
+size_t str_formatInt64(int64_t number, char* out, size_t outSize)
+{
+	ASSERT(outSize < BUFFER_SIZE_PARANOIA);
+
+	char scratchBuffer[30];
```
janmazak, line 106: I think I would rename `scratchBuffer` to better reflect that it is used not as a byte buffer, but string, e.g. `tmpStr`.

src/textUtils.c
```diff
@@ -59,39 +59,70 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
...
+		if (INT64_MIN == number) {
+			signlessNumber = ((uint64_t)INT64_MAX) + 1;
+		} else {
+			signlessNumber = (uint64_t) llabs(number);
+		}
+	} else {
+		signlessNumber = (uint64_t) number;
+	}
+	// Size without terminating character (and potential sign)
+	size_t rawSize = stringifyUint64ToBufferReverse(signlessNumber, ptr, end - 1);
```
janmazak, line 121: I think we mostly use "length" and not "size" when dealing with text strings. In particular, size should be length + 1.

src/textUtils.c
```diff
@@ -59,39 +59,70 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
...
+	return printedChars;
+}
 
-	// Size without terminating character
-	STATIC_ASSERT(sizeof(ptr - scratchBuffer) == sizeof(size_t), "bad size_t size");
-	size_t rawSize = (size_t) (ptr - scratchBuffer);
-
-	if (rawSize + 1 > outSize) {
+static void printReversedStringToBuffer(const char* reverted, size_t revertedSize, char* out, size_t outSize)
+{
```
janmazak, line 76: Shouldn't we assert that `strlen(reverted) == revertedSize`? Or remove the revertedSize argument altogether?

src/textUtils.c
```diff
@@ -59,39 +59,70 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
 	return rawSize + suffixLength;
 }
 
-size_t str_formatUint64(uint64_t number, char* out, size_t outSize)
+static size_t stringifyUint64ToBufferReverse(uint64_t number, char* ptr, char* end)
```
janmazak, line 62: Maybe more meaningful names for ptr and end so that it is clear what they mean?

src/textUtils.c
```diff
@@ -59,39 +59,70 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
...
+	char scratchBuffer[30];
+	char* ptr = BEGIN(scratchBuffer);
+	char* end = END(scratchBuffer);
+
+	uint64_t signlessNumber;
+	if (number < 0) {
+		if (INT64_MIN == number) {
+			signlessNumber = ((uint64_t)INT64_MAX) + 1;
+		} else {
+			signlessNumber = (uint64_t) llabs(number);
```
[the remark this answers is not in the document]
mkv-vcm: Or avoid llabs overall, and just write (uint64_t)(-number)?

src/textUtils.c
mkv-vcm: Can't really limit these in a sensible way, since they directly result in rawSizes, which is the return value in the end. Of course we assume that rawSizes is equal to strlen(out), so it can be done if we want it.

src/textUtils.c
mkv-vcm: Rawsize is superfluous if we just count the length again. But sure, I can put a 0 at the end of the scratch area, I just don't see the attack vector here. What _can_ go wrong?

src/textUtils.c
mkv-vcm: IF it's closed by a 0 then sure. Previously it was not, because why would it be, it's not a string to display, just a temporary working area.

src/textUtils.c
mkv-vcm: The WRITE_CHAR macro itself calls them ptr and end, and end is pretty self explanatory I think. I renamed ptr to currChar.

src/textUtils.c
janmazak: A string without \0 is a no-go if it can be avoided. Even if you cannot imagine a scenario where it would go wrong _right now_, sooner or later someone will modify your code or use your function for another purpose, and then in can go wrong easily (e.g. by allowing an attacker to read the memory where some data is stored he should not be able to access).

If a function writes into a string buffer and returns the length of the written string, one should be able to assert that `strlen(buffer) == returned_length`, and it's best to keep this invariant as much as possible even inside functions.

janmazak: Remember, we are somewhat paranoid about security here.

src/textUtils.c
```diff
@@ -89,41 +91,49 @@ size_t str_formatUint64(uint64_t number, char* out, size_t outSize)
 {
 	ASSERT(outSize < BUFFER_SIZE_PARANOIA);
 
-	char scratchBuffer[30];
-	char* ptr = BEGIN(scratchBuffer);
-	char* end = END(scratchBuffer);
+	{
+		char tmpReverted[30];
```
janmazak, line 95: `tmpReversed`? Also in `str_formatInt64`.

src/textUtils.c
```diff
@@ -89,41 +91,49 @@ size_t str_formatUint64(uint64_t number, char* out, size_t outSize)
...
+		char* begin = BEGIN(tmpReverted);
+		char* end = END(tmpReverted);
 
-	// Size without terminating character
-	const size_t rawSize = stringifyUint64ToBufferReverse(number, ptr, end);
-	printReversedStringToBuffer(scratchBuffer, rawSize, out, outSize);
-	return rawSize;
+		// Size without terminating character
+		stringifyUint64ToBufferReverse(number, begin, end);
+		const size_t revertedSize = strlen(tmpReverted);
```
janmazak, line 101: `reversedSize`?

src/textUtils.c
```diff
@@ -59,21 +59,23 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
 	return rawSize + suffixLength;
 }
 
-static size_t stringifyUint64ToBufferReverse(uint64_t number, char* ptr, char* end)
+static size_t stringifyUint64ToBufferReverse(uint64_t number, char* currChar, char* end)
```
janmazak, line 62: As I think about it, we generally pass buffer pointer and buffer size as arguments, e.g. for `str_formatAdaAmount`. I would prefer if this function conformed to this convention.
It's true you need end ptr for WRITE_CHAR, but you can compute it locally from the size.

src/textUtils.c
```diff
@@ -89,41 +91,49 @@ size_t str_formatUint64(uint64_t number, char* out, size_t outSize)
...
-		scratchBuffer[rawSize++] = '-';
+	{
+		char tmpReverted[30];
+		char* begin = BEGIN(tmpReverted);
+		char* end = END(tmpReverted);
+		// Size without terminating character (and potential sign)
+		stringifyUint64ToBufferReverse(signlessNumber, begin, end - 1);
+		size_t revertedSize = strlen(tmpReverted);
+		if (number < 0) {
+			snprintf(tmpReverted + revertedSize, sizeof(tmpReverted) - revertedSize, "%c", '-');
```
janmazak, line 130:
```suggestion
			snprintf(tmpReverted + revertedSize, SIZEOF(tmpReverted) - revertedSize, "%c", '-');
```

janmazak (commented):
PRs should be not be made against `master`, but against `develop` (or some other branch occasionally, but never master).

(One of the reasons is that Ledger maintains a fork of our repo where they sometimes commit some small changes of their own, then we need to update our master to reflect the changes and rebase develop on top of it.)

src/textUtils.c
```diff
@@ -75,32 +78,29 @@ static size_t stringifyUint64ToBufferReverse(uint64_t number, char* currChar, ch
 
 static void printReversedStringToBuffer(const char* reverted, char* out, size_t outSize)
```
janmazak, line 79: reverted -> reversed

src/textUtils.c
```diff
@@ -120,18 +120,16 @@ size_t str_formatInt64(int64_t number, char* out, size_t outSize)
...
-		char tmpReverted[30];
-		char* begin = BEGIN(tmpReverted);
-		char* end = END(tmpReverted);
-		// Size without terminating character (and potential sign)
-		stringifyUint64ToBufferReverse(signlessNumber, begin, end - 1);
-		size_t revertedSize = strlen(tmpReverted);
+		char tmpReversed[30];
+		//size without the potential '-' sign
+		stringifyUint64ToBufferReverse(signlessNumber, tmpReversed, SIZEOF(tmpReversed) - 1);
+		size_t reversedSize = strlen(tmpReversed);
```
janmazak, line 126: reversedSize -> reversedLength

janmazak (approved):
Should be merged as part of a PR that actually uses this code (mint).

janmazak:
Will be merged as part of another PR.

PR #76, opened 2021-05-20
add mint to txHashBuilder and device tests

mkv-vcm:
Some of the obvious changes will be reverted in the final form.


src/txHashBuilder.c
```diff
@@ -290,15 +293,37 @@ void txHashBuilder_addOutput_token(
 
 	if (builder->outputData.remainingTokens == 0) {
```
janmazak, line 294: `outputData` need to be renamed to `multiassetData` or something like that, and we should carefully check they are initialized properly.

src/txHashBuilder.c
```diff
@@ -257,19 +258,21 @@ void txHashBuilder_addOutput_tokenGroup(
...
-        const uint8_t* assetNameBuffer, size_t assetNameSize,
-        uint64_t amount
-)
+__noinline_due_to_stack__
+static void addToken(tx_hash_builder_t* builder,
+                     const uint8_t* assetNameBuffer, size_t assetNameSize,
+                     uint64_t amount,
+                     tx_hash_builder_state_t mandatoryState,
+                     tx_hash_builder_state_t nextGroupState,
+                     tx_hash_builder_state_t backoutState)
```
janmazak, line 271: Perhaps `leavingState`? I am afraid "backout" is not clear enough, or not common enough as a word.

src/txHashBuilder.c
```diff
@@ -862,9 +887,88 @@ static void txHashBuilder_assertCanLeaveWithdrawals(tx_hash_builder_t* builder)
 	ASSERT(builder->remainingWithdrawals == 0);
 }
 
-void txHashBuilder_addAuxData(tx_hash_builder_t* builder, const uint8_t* auxDataHashBuffer, size_t auxDataHashBufferSize)
+void txHashBuilder_enterMint(tx_hash_builder_t* builder)
 {
 	txHashBuilder_assertCanLeaveWithdrawals(builder);
```
janmazak, line 892: mint comes after validity interval start, not withdrawals?

src/txHashBuilder.c
```diff
@@ -862,9 +887,88 @@ static void txHashBuilder_assertCanLeaveWithdrawals(tx_hash_builder_t* builder)
...
+
+	builder->outputData.remainingAssetGroups = numAssetGroups;
+	// Map(numAssetGroups)[
+	//   { * policy_id => { * asset_name => uint } }
+	// ]
+	BUILDER_APPEND_CBOR(CBOR_TYPE_MAP, numAssetGroups);
+	if (0 != numAssetGroups) {
+		builder->state = TX_HASH_BUILDER_IN_MINT_ASSET_GROUP;
+	} else {
+		// TODO is this an error case?
```
janmazak, line 916: Generally, the tx hash builder is a bit more general than APDU handlers. I.e. it accepts and correctly hashes more than security policies and APDU validation allow. So if this is an error, it would be a VALIDATE error in some APDU handler, not an error here. We can leave a comment like "tx hash builder allows empty mint" --- AFAIK it is not forbidden to have empty mint on the protocol level and it would be accepted by a node, though I'll ask on slack to clarify it.

https://input-output-rnd.slack.com/archives/G010N9UFDNH/p1621934329014800

Please make sure that our implementation is consistent with their answer.

Right now, I have a suspicion that we do allow empty maps, but then the line
```
		ASSERT(subctx->numTokens > 0);
```
looks like we are asserting something that has not been validated before. That would be a bug.

src/txHashBuilder.c
```diff
@@ -862,9 +887,88 @@ static void txHashBuilder_assertCanLeaveWithdrawals(tx_hash_builder_t* builder)
...
+	              TX_HASH_BUILDER_IN_MINT_ASSET_GROUP, TX_HASH_BUILDER_IN_MINT_TOKEN);
+}
+
+void txHashBuilder_addMint_token(
+        tx_hash_builder_t* builder,
+        const uint8_t* assetNameBuffer, size_t assetNameSize,
+        int64_t amount
+)
+{
+	//  TODO int64 -> uint64 conversion should work fine ?
```
janmazak, line 936: Unfortunately not, if the int64 is too big (cannot be represented by uint64). There is no simple type that could hold both. The only way around it would be to have a union of the two plus one byte for an enum to distinguish how to interpret it, but that sounds like an overkill (I would not do it just for the purpose of txhashbuilder, but having `multiasset_amount_t` might actually be useful if we wanted to have common APDU handlers and/or ui state machines).

Also, one cannot use
`BUILDER_APPEND_CBOR(CBOR_TYPE_UNSIGNED, amount);` with int64.

Please consult with @balintklement  that ledgerjs should contain tests with maximum- and minimum-valued integers in multiasset outputs and mints.

src/txHashBuilder.c
```diff
@@ -862,9 +887,88 @@ static void txHashBuilder_assertCanLeaveWithdrawals(tx_hash_builder_t* builder)
...
+
+	switch (builder->state) {
+	case TX_HASH_BUILDER_IN_MINT:
+		break;
+
+	case TX_HASH_BUILDER_IN_WITHDRAWALS:
+	case TX_HASH_BUILDER_IN_CERTIFICATES:
+	case TX_HASH_BUILDER_IN_TTL:
+	case TX_HASH_BUILDER_IN_FEE:
+		txHashBuilder_assertCanLeaveWithdrawals(builder);
```
janmazak, line 957: These should be ordered consistently with other such lists (the state machine is more or less linear).

src/txHashBuilder.c
```diff
@@ -862,9 +887,88 @@ static void txHashBuilder_assertCanLeaveWithdrawals(tx_hash_builder_t* builder)
...
+	}
+
+	ASSERT(builder->outputData.remainingAssetGroups == 0);
+	ASSERT(builder->outputData.remainingTokens == 0);
+}
+
+
+void txHashBuilder_addAuxData(tx_hash_builder_t* builder, const uint8_t* auxDataHashBuffer, size_t auxDataHashBufferSize)
+{
+	txHashBuilder_assertCanLeaveMint(builder);
```
janmazak, line 971: Mint is the last item in transaction_body in 
https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl

src/txHashBuilder.h
```diff
@@ -12,6 +12,7 @@ enum {
 	TX_BODY_KEY_CERTIFICATES = 4,
 	TX_BODY_KEY_WITHDRAWALS = 5,
 	// TX_BODY_KEY_UPDATE = 6, // not used
+	TX_BODY_KEY_MINT = 9,   //TODO can these be rearranged? they go into the CBOR
```
janmazak, line 15: They cannot, we are computing a rolling hash.

janmazak:
PR should be created against the `develop` branch (or possibly some other, but never `master`).

mkv-vcm:
The [TEMP] testing commit is just there as a proof of concept that will be left out of the final commit; it produces the same hash in the end as the following cli command:

```
FIRST_POLICY_NAME="ad80794f72fa0edde3218b8d3fda801c24e25cbe61249e091d7eb10c"
SECOND_POLICY_NAME="dd56053690281363b13f4f885cb734e26ba0b1f197b093402c503f5e"
FIRST_ASSET_NAME="baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
SECOND_ASSET_NAME="caaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
THIRD_ASSET_NAME="daaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
FOURTH_ASSET_NAME="e"

cc transaction build-raw --tx-in "0B40265111D8BB3C3C608D95B3A0BF83461ACE32D79336579A1939B3AAD1C0B7#0" \
--tx-out "Ae2tdPwUPEZ7sZ29n6td4jkbnK31Q2BdQQ2rrMGYAXeEjVznrixjc5PZ2rj+100" \
--fee 42 --out-file tx.raw --mary-era \
--mint "110 $FIRST_POLICY_NAME.$FIRST_ASSET_NAME+120 $FIRST_POLICY_NAME.$SECOND_ASSET_NAME+210 $SECOND_POLICY_NAME.$THIRD_ASSET_NAME+220 $SECOND_POLICY_NAME.$FOURTH_ASSET_NAME" --minting-script-file ./policy.script

cc transaction txid --tx-body-file tx.raw
```
I still plan to include mint in the overall testing, but right now I want to move on to the transaction side.

mkv-vcm:
Included in the multisig PR.

PR #78, opened 2021-06-03
Mint transaction

mkv-vcm:
Mint transaction hashing.


src/txHashBuilder.c
```diff
@@ -1047,11 +1077,93 @@ static void txHashBuilder_assertCanLeaveValidityIntervalStart(tx_hash_builder_t*
...
+
+	builder->multiassetData.remainingAssetGroups = numAssetGroups;
+	// Map(numAssetGroups)[
+	//   { * policy_id => { * asset_name => uint } }
+	// ]
+	BUILDER_APPEND_CBOR(CBOR_TYPE_MAP, numAssetGroups);
+	if (0 != numAssetGroups) {
+		builder->state = TX_HASH_BUILDER_IN_MINT_ASSET_GROUP;
+	} else {
+		// TODO is this an error case?
```
janmazak, line 1108: I don't think there is any point in serializing empty asset groups. Choose the way that is simpler --- just serialize it correctly without validation (might work out of the box with empty cbor arrays), or ASSERT num > 0 and avoid serializing altogether if it would add addtional complexity. (I guess we can have the assertion in both cases anyway.)

src/txHashBuilder.c
```diff
@@ -1047,11 +1077,93 @@ static void txHashBuilder_assertCanLeaveValidityIntervalStart(tx_hash_builder_t*
...
+	              TX_HASH_BUILDER_IN_MINT_ASSET_GROUP, TX_HASH_BUILDER_IN_MINT_TOKEN);
+}
+
+void txHashBuilder_addMint_token(
+        tx_hash_builder_t* builder,
+        const uint8_t* assetNameBuffer, size_t assetNameSize,
+        int64_t amount
+)
+{
+	//  TODO int64 -> uint64 conversion should work fine ?
```
janmazak, line 1128: Only if the int is non-negative?

janmazak (commented):
The suggestions for code sharing with outputs might wait a bit I guess --- perhaps it makes sense to first have a working version, coupled with reasonably extensive ledgerjs tests, and then do smaller refactors while checking nothing got broken.

I would say `advanceCertificatesStateIfAppropriate` should be renamed to `advanceCertificatesStageIfAppropriate` (not related to your changes, perhaps just add a TODO).

src/securityPolicy.c
```diff
@@ -663,6 +663,31 @@ security_policy_t policyForSignTxValidityIntervalStart()
 	SHOW();
 }
 
+security_policy_t policyForSignTxMintDefault()
```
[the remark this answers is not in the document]
mkv-vcm: I don't even know what this function should do. I'm completely in the dark about how the security policies should work in minting.

src/signTxMint.c
```diff
@@ -0,0 +1,393 @@
...
+	advanceState();
+}
+
+enum {
+	HANDLE_ASSET_GROUP_STEP_DISPLAY = 800, // TODO
+	HANDLE_ASSET_GROUP_STEP_RESPOND,
+	HANDLE_ASSET_GROUP_STEP_INVALID,
+};
+
+static void signTxMint_handleAssetGroup_ui_runStep()
```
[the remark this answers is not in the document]
mkv-vcm: The same.

src/signTxMint.c
```diff
@@ -0,0 +1,393 @@
...
+		VALIDATE(numAssetGroups <= OUTPUT_ASSET_GROUPS_MAX, ERR_INVALID_DATA);
+
+		STATIC_ASSERT(OUTPUT_ASSET_GROUPS_MAX <= UINT16_MAX, "wrong max token groups");
+		ASSERT_TYPE(subctx->numAssetGroups, uint16_t);
+		subctx->numAssetGroups = (uint16_t) numAssetGroups;
+
+		VALIDATE(view_remainingSize(&view) == 0, ERR_INVALID_DATA);
+	}
+	txHashBuilder_addMint_topLevelData(txHashBuilder, subctx->numAssetGroups);
+	subctx->mintSecurityPolicy = policyForSignTxMintDefault();	//TODO not sure how to proceed here
```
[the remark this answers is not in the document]
mkv-vcm: I don't understand that. If not kept in a global state, then where do we need a value at all? Right now it's set in one place and used in two others, but that was just a copy of Outputs.

src/txHashBuilder.c
mkv-vcm: I think this describes it pretty well. The signed is simply interpreted as an unsigned, and that's just fine for us.
https://stackoverflow.com/questions/50605/signed-to-unsigned-conversion-in-c-is-it-always-safe

src/securityPolicy.c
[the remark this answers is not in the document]
janmazak: In general, there are some security policies that do nothing at present (just allow to proceed).

For minting, the policy should take signing mode as an argument and decide if minting is allowed or not: since it is governed by a script, it should be allowed in multisig txs and forbidden in all the rest (ordinary, pool reg.).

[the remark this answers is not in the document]
janmazak: At present, that's all we want from the policy AFAIK.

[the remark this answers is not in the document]
mkv-vcm: Where do I get the signing mode? Should that be the txUsecase in common_tx_data_t, just not implemented yet?

src/signTxMint.c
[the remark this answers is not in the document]
janmazak: Usually, the policy is used to decide the starting step for a ui state machine; a local variable to store the policy is enough for that. For outputs, exceptionally, we apply the policy for individual tokens, too (therefore we need a variable in global context to store it between APDUs). For mint, we do need that, because we are always going to show it, but since the global context uses unions, it probably does not affect total context size at all (please check). So I would prefer to keep the code as close to outputs as possible, i.e. having the global variable even if it was actually needed only here.

src/txHashBuilder.c
janmazak: I don't understand. A negative int64 must not be serialized as `CBOR_TYPE_UNSIGNED`, it is `CBOR_TYPE_NEGATIVE`. This is how we are supposed to serialize integers:
https://datatracker.ietf.org/doc/html/rfc7049#section-3.9
If they are too big (I'm not sure now), we might need something like
https://datatracker.ietf.org/doc/html/rfc7049#section-2.4.2

mkv-vcm: I see what you mean, right now negative values always take up the maximum 8 bytes, since as unsigned they are always larger than VALUE_MIN_W8. I'm not clear on who will read this CBOR (and the hash made from it) and who we want to be consistent with. I'll implement the negative side of CBOR.

janmazak: It would be great if we had examples of transactions with negative mint values, including ones that do not fit in uint32. Cardano-cli might actually not be complying with the CBOR spec, as we have discovered for token ordering. Please consult this with @gabrielKerekes who probably already has such txs.

mkv-vcm: LedgerJS already tests all the way to INT64_MIN, but I don't remember if my manual test with the generated hash value of minting was about a negative amount or not. But I think the more interesting case is where it _does_ fit into 32 bit, since that's where a difference should come up.

src/cbor.c
```diff
@@ -112,6 +126,16 @@ size_t cbor_writeToken(uint8_t type, uint64_t value, uint8_t* buffer, size_t buf
 
 	// Check sanity
 	switch (type) {
+	case CBOR_TYPE_NEGATIVE: {
+		int64_t negativeValue;
+		// reinterpret an actually negative value hidden in an unsigned in the safe way
+		memcpy(&negativeValue, &value, SIZEOF(value));
```
janmazak, line 132: I guess we want some `STATIC_ASSERT` before this, e.g. `SIZEOF(value) == SIZEOF(negativeValue)`.

src/cbor.c
```diff
@@ -45,49 +46,62 @@ cbor_token_t cbor_parseToken(const uint8_t* buf, size_t size)
...
+		if (result.value < limit_min) {
+			// This isn't canonical CBOR
+			THROW(ERR_UNEXPECTED_TOKEN);
+		}
 	}
 
-	if (result.value < limit_min) {
-		// This isn't canonical CBOR
-		THROW(ERR_UNEXPECTED_TOKEN);
+	if (CBOR_TYPE_NEGATIVE == result.type) {
```
janmazak, line 94: I think we don't need to parse CBOR_TYPE_NEGATIVE. It is OK to have it here, but it should probably be just commented out (to avoid increasing app size) if we ever needed it in the future. (AFAIK we are only parsing CBOR for some legacy Byron stuff, like extracting protocol magic, or something like that.) But the size difference is perhaps negligible, so I guess I am OK with having it here (especially if we can't easily comment it out in a single place, I've not really checked).

mkv-vcm:
Included in the multisig PR, except for the possible code size reduction by combining output and mint code, but that's more complicated than it seems at first glance.

PR #79, opened 2021-06-03
fix output ui

janmazak:
A quick fix of the problem of not displaying address parameters when they should be.

But we should rethink the UI a bit; the screen with the first line "WARNING" for enterprise addresses looks too disquieting when it appears out of nowhere in a tx output for user's own enterprise address which could be entirely his intention. So I suggest that even for addresses that can be decomposed into params we first show the bech32-encoded address and then its components.


janmazak:
Merged at Ledger repo.

PR #80, opened 2021-06-09
MultiSig

mkv-vcm:
@janmazak, I'm struggling here a lot with the definitions of what multisig paths should be (are 3 length multisig paths accepted in some way?) and where they should be handled how. Put some TODOs in the code, but I think the code speaks for itself about my uncertainties.


src/securityPolicy.c
```diff
@@ -103,6 +103,7 @@ security_policy_t policyForGetExtendedPublicKey(const bip44_path_t* pathSpec)
 		break;
 
 	default:
+		// TODO should multisig keys be allowed here?
```
janmazak, line 106: I think we should treat multisig keys in the same way as ordinary keys here (the potential differences are hidden in `bip44_isPathReasonable`).

src/securityPolicy.c
```diff
@@ -134,6 +135,7 @@ security_policy_t policyForGetExtendedPublicKeyBulkExport(const bip44_path_t* pa
...
```
janmazak, line 138: Again, I think we should do the same as with ordinary keys.

I would use `PATH_MULTISIG_*` instead of `PATH_WALLET_*` in the path classification enum for multisig keys.

janmazak (commented):
In key derivation path validations, notice that `role` cannot be 1 for multisig keys.
https://cips.cardano.org/cips/cip1854/
https://cips.cardano.org/cips/cip1852/
We need to make sure we are compliant with both of these.

src/addressUtilsShelley.c
```diff
@@ -22,11 +22,17 @@ address_type_t getAddressType(uint8_t addressHeader)
 bool isSupportedAddressType(uint8_t addressType)
 {
 	switch (addressType) {
-	case BASE:
-	case POINTER:
-	case ENTERPRISE:
+	case BASE_STAKE_KEY_PAYMENT_KEY:
```
janmazak, line 25: Payment part should be mentioned first.

src/addressUtilsShelley.h
```diff
@@ -7,13 +7,26 @@
...
+	REWARD_KEY							= 0b1110,
+	REWARD_SCRIPT						= 0b1111,
+
 	BASE       = 0b0000,  // 0x0
 	POINTER    = 0b0100,  // 0x4
 	ENTERPRISE = 0b0110,  // 0x6
-	BYRON      = 0b1000,  // 0x8
+	// BYRON      = 0b1000,  // 0x8
 	REWARD     = 0b1110,  // 0xE
+	MULTISIG   = 0b1111,  // 0xF TODO
```
janmazak, line 29: These seem superfluous.

src/bip44.c
```diff
@@ -72,8 +72,20 @@ bool bip44_hasShelleyPrefix(const bip44_path_t* pathSpec)
...
+{
+#define CHECK(cond) if (!(cond)) return false
+	CHECK(pathSpec->length > BIP44_I_COIN_TYPE);
+	CHECK(pathSpec->path[BIP44_I_PURPOSE] == (PURPOSE_MULTISIG | HARDENED_BIP32));
+	CHECK(pathSpec->path[BIP44_I_COIN_TYPE] == (ADA_COIN_TYPE | HARDENED_BIP32));
+	return true;
+#undef CHECK
+}
+
 bool bip44_hasValidCardanoWalletPrefix(const bip44_path_t* pathSpec)
```
janmazak, line 85: Some renaming is in order. My suggestion would be something like this:
CardanoWalletPrefix -> OrdinaryWalletPrefix
MultisigPrefix -> MultisigWalletPrefix
bip44_isValidStakingKeyPath -> bip44_isOrdinaryStakingKeyPath
analogously we are likely to need bip44_isMultisigStakingKeyPath (we are not going to mix the two since tx singing mode implies one or the other)

src/bip44.c
```diff
@@ -247,6 +259,36 @@ size_t bip44_printToStr(const bip44_path_t* pathSpec, char* out, size_t outSize)
 	return ptr - out;
 }
 
+static bip44_path_type_t bip44_classifyCardanoWalletPath(const bip44_path_t* pathSpec)
```
janmazak, line 262:
```suggestion
static bip44_path_type_t bip44_classifyWalletPath(const bip44_path_t* pathSpec)
```

src/bip44.c
```diff
@@ -247,6 +259,36 @@ size_t bip44_printToStr(const bip44_path_t* pathSpec, char* out, size_t outSize)
...
+	case 3: {
+		return multiSig ? PATH_MULTISIG_ACCOUNT : PATH_WALLET_ACCOUNT;
+	}
+	case 5: {
+		const uint8_t chainType = bip44_getChainTypeValue(pathSpec);
+		switch (chainType) {
+
+		case CARDANO_CHAIN_INTERNAL:
+		case CARDANO_CHAIN_EXTERNAL:
+			return multiSig ? PATH_MULTISIG_SPENDING_KEY : PATH_WALLET_SPENDING_KEY;
```
janmazak, line 275:
```suggestion
			return multiSig ? PATH_MULTISIG_SPENDING_KEY : PATH_ORDINARY_SPENDING_KEY;
```

janmazak: Multisig does not allow CARDANO_CHAIN_EXTERNAL afaik.

src/bip44.c
```diff
@@ -258,37 +300,8 @@ bip44_path_type_t bip44_classifyPath(const bip44_path_t* pathSpec)
 		}
```
janmazak, line 300: It perhaps makes sense to extract all 1853-related stuff into a separate function, e.g. `bip44_classifyPathPoolColdKey`.

src/bip44.h
```diff
@@ -46,6 +47,7 @@ enum {
 
 bool bip44_hasByronPrefix(const bip44_path_t* pathSpec);
 bool bip44_hasShelleyPrefix(const bip44_path_t* pathSpec);
+bool bip44_hasMultisigPrefix(const bip44_path_t* pathSpec);
 bool bip44_hasValidCardanoWalletPrefix(const bip44_path_t* pathSpec);
```
janmazak, line 51:
```suggestion
bool bip44_hasCardanoWalletPrefix(const bip44_path_t* pathSpec);
```

src/bip44.c
```diff
@@ -72,8 +72,20 @@ bool bip44_hasShelleyPrefix(const bip44_path_t* pathSpec)
...
+	CHECK(pathSpec->length > BIP44_I_COIN_TYPE);
+	CHECK(pathSpec->path[BIP44_I_PURPOSE] == (PURPOSE_MULTISIG | HARDENED_BIP32));
+	CHECK(pathSpec->path[BIP44_I_COIN_TYPE] == (ADA_COIN_TYPE | HARDENED_BIP32));
+	return true;
+#undef CHECK
+}
+
 bool bip44_hasValidCardanoWalletPrefix(const bip44_path_t* pathSpec)
 {
+	// TODO should bip44_hasMultisigPrefix come here? does it fulfill the
```
janmazak, line 87: This function is used very rarely. I would consider removing it altogether, but not in case it would lead to annoyingly long if/assertion conditions.  I'd remove "Valid" from the name and include multisig here (at least it seems to me it is used that way).

src/uiScreens.c
```diff
@@ -350,11 +350,11 @@ void ui_displayStakingInfoScreen(
 			heading = STAKING_HEADING_WARNING;
 			strncpy(stakingInfo, "legacy Byron address (no staking rewards)", SIZEOF(stakingInfo));
 
-		} else if (addressParams->type == ENTERPRISE) {
+		} else if (addressParams->type == ENTERPRISE_KEY || addressParams->type == ENTERPRISE_SCRIPT) {
```
janmazak, line 353: This long if looks like a good candidate for refactoring into a switch statement.

PR #85, opened 2021-07-15
Add native script derivation

KubqoA:


src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+#	define  CASE(TYPE, HANDLER) case TYPE: HANDLER(view); break;
+		CASE(PUBKEY_DEVICE_OWNED, deriveNativeScriptHash_handleDeviceOwnedPubkey);
+		CASE(PUBKEY_THIRD_PARTY, deriveNativeScriptHash_handleThirdPartyPubkey);
+		// any other value for the pubkey type is invalid
+		default: THROW(ERR_INVALID_DATA);
+#	undef   CASE
+	}
+}
+
+#define SPLIT_UINT64(n) (uint32_t) (n >> 32), (uint32_t) n
```
janmazak, line 222: Variables should be always parenthesized in macro definitions.

But I'm not sure if we need this splitting at all. I guess I would prefer `TRACE_UINT64` here.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+	VALIDATE(view_remainingSize(view) == ADDRESS_KEY_HASH_LENGTH, ERR_INVALID_DATA);
+	view_memmove(hash, view, ADDRESS_KEY_HASH_LENGTH);
+
+#	ifdef DEVEL
+	TRACE("Pubkey given by hash:");
+	for (int i = 0; i < ADDRESS_KEY_HASH_LENGTH; i++) {
+		PRINTF("%x", hash[i]);
+	}
+	PRINTF("\n");
+#	endif
```
janmazak, line 197: Use `TRACE_BUFFER`.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+
+	ui_displayHexBufferScreen(
```
janmazak, line 199: Don't we use bech32 encoding for displaying keys?

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+	TRACE("pubkey type = %u", pubkeyType);
+
+	switch(pubkeyType) {
+#	define  CASE(TYPE, HANDLER) case TYPE: HANDLER(view); break;
+		CASE(PUBKEY_DEVICE_OWNED, deriveNativeScriptHash_handleDeviceOwnedPubkey);
+		CASE(PUBKEY_THIRD_PARTY, deriveNativeScriptHash_handleThirdPartyPubkey);
+		// any other value for the pubkey type is invalid
+		default: THROW(ERR_INVALID_DATA);
+#	undef   CASE
+	}
```
janmazak, line 219: I guess I would write this explicitly without a macro, with direct calls in switch cases. Hmm. Probably would be easier to audit; the macro does not add much value (makes the code longer and more obscure?).

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+	uint8_t nativeScriptType = parse_u1be(view);
+	TRACE("native complex script type = %u", nativeScriptType);
+
+	switch(nativeScriptType) {
+#	define  CASE(TYPE, HANDLER) case TYPE: HANDLER(view); break;
+		CASE(NATIVE_SCRIPT_ALL, deriveNativeScriptHash_handleAll);
+		CASE(NATIVE_SCRIPT_ANY, deriveNativeScriptHash_handleAny);
+		CASE(NATIVE_SCRIPT_N_OF_K, deriveNativeScriptHash_handleNofK);
+		default: THROW(ERR_INVALID_DATA);
+#	undef   CASE
```
janmazak, line 162: Should be done even before `default`.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+	VALIDATE(expectsRequest(), ERR_INVALID_STATE);
+
+	// check if we can increase the level without breaking the MAX_SCRIPT_DEPTH constraint
+	VALIDATE(ctx->level + 1 < MAX_SCRIPT_DEPTH, ERR_DATA_TOO_LARGE);
+	ctx->level++;
+
+	uint8_t nativeScriptType = parse_u1be(view);
+	TRACE("native complex script type = %u", nativeScriptType);
+
+	switch(nativeScriptType) {
```
janmazak, line 156: Does this conform to our astyle rules? Try `make format`.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+		"Script - ANY",
+		text,
+		&deriveScriptHash_display_ui_callback
+	);
+}
+
+void deriveNativeScriptHash_handleNofK(read_view_t* view)
+{
+	// parse data
+	ctx->nativeScript.nOfK.n = parse_u4be(view);
```
janmazak, line 119: Some explicit validation should be done before attempting to read the view (in other places, too).

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// UI
+
+static void deriveScriptHash_display_ui_callback()
+{
+	if (shouldFinish()) {
+		TRACE("whole native script received");
+		ui_idle();
+		// TODO: hashing, this is just a mock response
```
janmazak, line 59: Hashing is needed after each APDU, not just once in the end. Generally, we first do computations (e.g. hashing) and only afterwards call ui operations.

janmazak (commented):
OK, looks rather good, life is simpler here since we don't need any security policies etc.

src/deriveNativeScriptHash.c
KubqoA: Didn't realize we had this macro as well, thanks

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+enum {
+	PUBKEY_DEVICE_OWNED = 0,
+	PUBKEY_THIRD_PARTY = 1,
+};
+
+static ins_derive_native_script_hash_context_t* ctx = &(instructionState.deriveNativeScriptHashContext);
+
+// Helper functions
+
+bool expectsRequest()
```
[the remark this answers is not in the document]
KubqoA: > All local functions should be declared `static` --- they are not visible outside of the module then.

Does this mean that also the `deriveNativeScriptHash_handle*` functions (aside from `handleAPDU`) should also be `static`?

src/deriveNativeScriptHash.c
KubqoA: Isn't it enough that the `parse_u?be` functions do this validation before parsing?
```
		VALIDATE(view_remainingSize(view) >= width, ERR_NOT_ENOUGH_INPUT); \
```

src/deriveNativeScriptHash.c
KubqoA: This is just a style preference, or is there something more to it?

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+
+static inline void scriptFinished()
+{
+	--ctx->remainingScripts[ctx->level];
+	TRACE("native script finished, level = %u, remainingScripts = %u", ctx->level, ctx->remainingScripts[ctx->level]);
+}
+
+static inline void complexScriptFinished()
+{
+	--ctx->level;
```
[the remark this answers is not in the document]
KubqoA: Okay, the two separate functions now explicitly do all the required operations and do some additional assertions.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+#include "uiScreens.h"
+
+enum nativeScriptType {
+	NATIVE_SCRIPT_PUBKEY = 0,
+	NATIVE_SCRIPT_ALL = 1,
+	NATIVE_SCRIPT_ANY = 2,
+	NATIVE_SCRIPT_N_OF_K = 3,
+	NATIVE_SCRIPT_INVALID_BEFORE = 4,
+	NATIVE_SCRIPT_INVALID_HEREAFTER = 5,
+};
```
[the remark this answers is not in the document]
KubqoA: Okay, moved it to the `deriveNativeScriptHash.h`

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+
+bool expectsRequest()
+{
+	// if the number of remaining scripts is not bigger than 0, then this request
+	// is invalid in the current context, as Ledger was not expecting another
+	// script to be parsed
+	return ctx->remainingScripts[ctx->level] > 0;
+}
+
+bool shouldFinish()
```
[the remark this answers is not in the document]
KubqoA: Renamed to `isWholeNativeScriptFinished`. Hopefully it is explicit enough about what is finished.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,341 @@
...
+{
+	// expect no data
+	VALIDATE(read_view_remaining_size(view) == 0, ERR_INVALID_DATA);
+
+	// if the number of remaining scripts for the current complex script is not 0
+	// then this request is invalid in the current context
+	VALIDATE(ctx->remainingScripts[ctx->level] == 0, ERR_INVALID_STATE);
+
+	switch(ctx->type[ctx->level]) {
+#	define  CASE(TYPE, HEADER) case TYPE: ui_displayPaginatedText(HEADER, "Finished", &deriveScriptHash_display_ui_callback); break;
```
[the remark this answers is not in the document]
KubqoA: Does this mean that also other references in the file to `&deriveScriptHash_display_ui_callback` should be changed to `deriveScriptHash_display_ui_callback`?

Also if I may ask, what's the difference? Because looking at the code I am a little bit confused. For example calling the `ui_displayUint64Screen` accepts `ui_callback_fn_t callback`:
```c
void ui_displayUint64Screen(
        const char* firstLine,
        uint64_t value,
        ui_callback_fn_t callback
)
{
	char valueStr[30];
	str_formatUint64(value, valueStr, SIZEOF(valueStr));

	ui_displayPaginatedText(
	        firstLine,
	        valueStr,
	        callback
	);
}
```
but the `ui_displayPaginatedText` has this signature:
```c
void ui_displayPaginatedText(
        const char* headerStr,
        const char* bodyStr,
        ui_callback_fn_t* callback)
```
namely the `callback` here is a pointer, but in the `ui_displayUint64Screen` the original `callback` is passed as it is. So what's going on here?

To me it seems like `ui_callback_fn_t*` and `ui_callback_fn_t` are the same, but I am probably missing something

src/deriveNativeScriptHash.c
KubqoA: I switched to bech32, but wasn't sure what prefix to use, tried to look for other usecases in the codebase, but that didn't help. For now went with the `pubkey` prefix, but I'll welcome any suggestions.

doc/ins_derive_native_script_hash.md
```diff
@@ -0,0 +1,97 @@
...
+| P1    | `0x01` |
+
+*Data*
+| Field | Length | Comments |
+|-------|--------|----------|
+| script type | 1 | Script type according to the [cddl](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L241) |
+| number of required scripts | 4 | Only when script type is `N_OF_K` |
+| number of scripts | 4 | |
+
+The *number_of_required_scripts* must be omitted for script types `ALL` and `ANY`.
```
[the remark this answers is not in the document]
KubqoA: Makes sense. Just to make sure, you propose I should change this then also in ledgerjs and ledger app right?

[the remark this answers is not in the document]
KubqoA: My original thinking when implementing this, was that the cddl defines the `n_of_k` script as follows:
```
script_n_of_k = (3, n: uint, [ * native_script ])
```
so I thought it would be okay, to have the number of required scripts defined before the number of the subscripts.

src/deriveNativeScriptHash.c
janmazak: Perhaps a bit of style, too, but I think the reason is to minimize the scope (for such a "local" macro). Notice how we are using {} to shorten scopes for variables, or even in places where they just group some function calls. We are kind of paranoid about security in this repo. Because that's the whole point of having a hw wallet.

src/deriveNativeScriptHash.c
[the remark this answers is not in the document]
janmazak: In C, the function name is also a pointer to the function. It is common to use it in that way, e.g. when you have a function pointer f, you usually call the function `f()`, not `(*f)()`, though both work. Similarly, you pass a function g() as either `g` or `&g` as parameter, and our codebase uses the first way, so we want to be consistent.
https://stackoverflow.com/questions/12152167/why-is-using-the-function-name-as-a-function-pointer-equivalent-to-applying-the

src/deriveNativeScriptHash.c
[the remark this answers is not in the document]
janmazak: I guess there is no strict rule on this; afaik using static helps compiler/linker with optimization (when creating an object module for a single .c file, it is not yet clear whether the function is being called from elsewhere or not; with `static`, it is forbidden to call it from elsewhere). And you avoid the problem of implicitly declared functions (not put in header, but available to linker) which only generates a warnings which is sometimes ignored. So I don't see reasons not to make a function static if it can be.

src/deriveNativeScriptHash.c
janmazak: This should be dealt with across the whole codebase. A very reasonable task for you, btw.; consult with Gabo or on the small daily standup the individual instances where we are displaying keys. The bech32 prefixes are here:
https://github.com/cardano-foundation/CIPs/blob/master/CIP-0005/CIP-0005.md

src/deriveNativeScriptHash.c
janmazak: The convention in this code base is that whenever you are going to read from a view, you check that the view actually contains the data (check the buffer length). As I think about it now, however, it cannot be reasonably done for instance for bip44 paths (variable length, checked within the function), so it cannot be absolute. For security, it is better to check everything twice than risk not checking at all, but since we have problem with total code size, it seems preferable to drop checks that are obviously superfluous and just rely on the functions reading the view. You should create a separate PR for this: identify all situations like
```
		VALIDATE(view_remainingSize(view) >= 4, ERR_INVALID_DATA);
		params->protocolMagic = parse_u4be(view);

```
and remove the validation.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,363 @@
...
+	ASSERT(ctx->complexScripts[ctx->level].remainingScripts > 0);
+	ctx->complexScripts[ctx->level].remainingScripts--;
+
+	TRACE_WITH_CTX("complex script finished, ");
+}
+
+// UI
+
+static void deriveScriptHash_display_ui_callback() {
+	io_send_buf(SUCCESS, NULL, 0);
```
janmazak, line 46: Take a look at `respondSuccessEmptyMsg()`. For some reasons that go fairly deep into very ledger-specific code, and I don't understand well, after io_send_buf, you want to also tell the UI something, e.g ui_displayBusy().

src/deriveNativeScriptHash.c
KubqoA: Added fix commits which apply the formatting as defined by `make format`.

KubqoA: Now everything seems to be okay according to `make format`

src/deriveNativeScriptHash.c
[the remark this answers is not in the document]
KubqoA: Okay, where it was possible I changed the function to `static`.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,520 @@
...
+
+	UI_STEP(DISPLAY_UI_STEP_POSITION) {
+		// For complex scripts we reduce the current level by 1
+		// Because they already have the level increased by 1
+		int levelOffset = ctx->ui_scriptType == UI_SCRIPT_ALL
+		                  || ctx->ui_scriptType == UI_SCRIPT_ANY
+		                  || ctx->ui_scriptType == UI_SCRIPT_N_OF_K
+		                  ? -1 : 0;
+		uint8_t level = ctx->level + levelOffset;
+		// if we are on the root level don't show the position
```
janmazak, line 148: I'd wrap this logic into a function `_shouldShowScriptPosition`.

src/deriveNativeScriptHash.h
```diff
@@ -0,0 +1,52 @@
...
+// to handle PUBKEY DEVICE_OWNED vs THIRD_PARTY
+typedef enum {
+	UI_SCRIPT_PUBKEY_PATH = 0,
+	UI_SCRIPT_PUBKEY_HASH,
+	UI_SCRIPT_ALL,
+	UI_SCRIPT_ANY,
+	UI_SCRIPT_N_OF_K,
+	UI_SCRIPT_INVALID_BEFORE,
+	UI_SCRIPT_INVALID_HEREAFTER,
+} ui_native_script_type;
```
janmazak, line 22: I'm not sure we really need this. Wouldn't a single if for ui_script_pubkey disctinction solve it?

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,520 @@
...
+
+	ctx->complexScripts[ctx->level].ui_type = UI_SCRIPT_N_OF_K;
+	UI_DISPLAY_SCRIPT(UI_SCRIPT_N_OF_K);
+}
+
+static void deriveNativeScriptHash_handleComplexScriptStart(read_view_t* view)
+{
+	TRACE_BUFFER(view->begin, (size_t) (view->end - view->begin));
+
+	expectsMoreScripts();
```
janmazak, line 275: This is confusing. It will be much clearer if there was a bool function, say, `areMoreScriptsExpected`, and then
`VALIDATE(areMoreScriptsExpected() == true, ERR_INVALID_STATE)` here.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,520 @@
...
+
+	// check if we can increase the level without breaking the MAX_SCRIPT_DEPTH constraint
+	VALIDATE(ctx->level + 1 < MAX_SCRIPT_DEPTH, ERR_DATA_TOO_LARGE);
```
janmazak, line 278: The error ERR_DATA_TOO_LARGE is not used consistently throughout the codebase, I'm afraid. The original intention is to report a buffer that is too large; I would suggest using ERR_INVALID_DATA here. (And please create an issue for the code cleanup regarding these two errors: they should get a clearly defined meaning, mentioned in a comment near their definition, and then used throughout the code according to the meaning.)

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,520 @@
...
+	ctx->level++;
+	explicit_bzero(&ctx->complexScripts[ctx->level], SIZEOF(ctx->complexScripts[ctx->level]));
+
+	uint8_t nativeScriptType = parse_u1be(view);
+	TRACE("native complex script type = %u", nativeScriptType);
+
+	switch(nativeScriptType) {
+#	define  CASE(TYPE, HANDLER) case TYPE:\
+		ctx->complexScripts[ctx->level].remainingScripts = parse_u4be(view);\
+		ctx->complexScripts[ctx->level].totalScripts = ctx->complexScripts[ctx->level].remainingScripts;\
```
janmazak, line 288: These two lines are independent of the CASE macro arguments, that's very fishy. Can't they be moved before the switch?

janmazak (commented):
I've not read through everything (especially not the UI machine code, but that needs a considerable rewrite anyway).

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,520 @@
...
+
+	nativeScriptHashBuilder_startComplexScript_n_of_k(&ctx->hashBuilder, ctx->scriptContent.requiredScripts, ctx->complexScripts[ctx->level].remainingScripts);
+
+	ctx->complexScripts[ctx->level].ui_type = UI_SCRIPT_N_OF_K;
+	UI_DISPLAY_SCRIPT(UI_SCRIPT_N_OF_K);
+}
+
+static void deriveNativeScriptHash_handleComplexScriptStart(read_view_t* view)
+{
+	TRACE_BUFFER(view->begin, (size_t) (view->end - view->begin));
```
janmazak, line 273: There is a macro for this purpose: `VIEW_REMAINING_TO_TUPLE_BUF_SIZE`.

(Should be applied several times in the code.)

src/nativeScriptHashBuilder.c
```diff
@@ -0,0 +1,186 @@
...
+
+#ifdef TRACE_NATIVE_SCRIPT_HASH_BUILDER
+#define _TRACE(...) TRACE(__VA_ARGS__)
+#define _TRACE_BUFFER(...) TRACE_BUFFER(__VA_ARGS__)
+#else
+#define _TRACE(...)
+#define _TRACE_BUFFER(...)
+#endif // TRACE_NATIVE_SCRIPT_HASH_BUILDER
+
+#define APPEND_CBOR(type, value) blake2b_224_append_cbor_data(&builder->nativeScriptHash, type, value);
```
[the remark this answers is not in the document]
KubqoA: Why is it better to define the macro without the semicolon at the end?

src/nativeScriptHashBuilder.c
```diff
@@ -0,0 +1,186 @@
...
+	ASSERT(builder->state == NATIVE_SCRIPT_HASH_BUILDER_SCRIPT);
+	\
+
+	APPEND_CBOR(CBOR_TYPE_ARRAY, 3);
+	APPEND_CBOR(CBOR_TYPE_UNSIGNED, NATIVE_SCRIPT_N_OF_K);
+	APPEND_CBOR(CBOR_TYPE_UNSIGNED, requiredScripts);
+	APPEND_CBOR(CBOR_TYPE_ARRAY, remainingScripts);
+
+	builder->level++;
+	builder->remainingScripts[builder->level] = remainingScripts;
```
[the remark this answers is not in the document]
KubqoA: Yes, we support empty nested scripts. There are also test cases for it directly on Ledger and also in Ledgerjs.

src/nativeScriptHashBuilder.h
```diff
@@ -0,0 +1,68 @@
+#ifndef H_CARDANO_APP_NATIVE_SCRIPT_HASH_BUILDER
+#define H_CARDANO_APP_NATIVE_SCRIPT_HASH_BUILDER
+
+#include "cardano.h"
+#include "cbor.h"
```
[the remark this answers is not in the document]
KubqoA: If you are referring to the `#include "cbor.h"`, then yes, it's not included when including `cardano.h`

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,520 @@
...
+		deriveNativeScriptHash_handleThirdPartyPubkey(view);
+		break;
+	// any other value for the pubkey type is invalid
+	default:
+		THROW(ERR_INVALID_DATA);
+	}
+}
+
+// TRACE doesn't handle uint64_t so we split it into two uint32_t
+#define TRACE_UINT64(varname, n) TRACE(varname "= 0x%x%x", (uint32_t) (n >> 32), (uint32_t) n)
```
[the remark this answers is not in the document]
KubqoA: Ah, now I see, it's not a macro, it's a function called `str_traceUint64`, which honestly seems much more complicated then it's needed for tracing a uint64. Do you want me to change it?
In the function call the uint64 is first converted to a string and then traced. My macro is in my opinion much more simpler for debug purposes.

src/nativeScriptHashBuilder.c
```diff
@@ -0,0 +1,186 @@
...
+}
+
+void nativeScriptHashBuilder_init(
+        native_script_hash_builder_t* builder
+)
+{
+	TRACE("Serializing native script hash data");
+	blake2b_224_init(&builder->nativeScriptHash);
+
+	// set the initial 0x00 byte
```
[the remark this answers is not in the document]
KubqoA: As the only relevant piece of documentation is the google docs, I just copied the relevant information directly to the comment.

src/nativeScriptHashBuilder.c
```diff
@@ -0,0 +1,186 @@
...
+void nativeScriptHashBuilder_finishComplexScript(
+        native_script_hash_builder_t* builder
+)
+{
+	_TRACE("state = %d", builder->state);
+
+	ASSERT(builder->state == NATIVE_SCRIPT_HASH_BUILDER_COMPLEX_SCRIPT_FINISHED);
+
+	builder->level--;
+	builder->remainingScripts[builder->level]--;
```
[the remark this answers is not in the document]
KubqoA: After removing the complex script finished call, the level is only decreased after checking whether the level is bigger than 0, so I added an assertion only for the second one.

src/nativeScriptHashBuilder.c
```diff
@@ -0,0 +1,186 @@
...
+	_TRACE_BUFFER(buffer, size);
+	blake2b_224_append(hashCtx, buffer, size);
+}
+
+static inline void advanceState(native_script_hash_builder_t* builder)
+{
+	_TRACE("Advancing state, level = %u, remaining scripts = %u", builder->level, builder->remainingScripts[builder->level]);
+	// if the current number of remaining scripts is bigger than 0 we still
+	// expect scripts
+	if (builder->remainingScripts[builder->level] > 0) {
```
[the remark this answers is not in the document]
KubqoA: I am not sure I understand. Currently the state has only two options in which it can be:
`NATIVE_SCRIPT_HASH_BUILDER_SCRIPT ` and `NATIVE_SCRIPT_HASH_BUILDER_FINISHED `. So in my opinion the switch statement you described doesn't make sense.
I added an assertion to check whether the state is `NATIVE_SCRIPT_HASH_BUILDER_SCRIPT `, because that's the only state this `advanceState` method should be called with. Is that okay with you?

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,520 @@
...
+		ui_displayPaginatedText(
+		        HEADER,
+		        "Finished",
+		        callback
+		);
+		return;
+	}
+
+	// "Position: x.x.x - Finished"
+	char content[10 + 11 * level + 10 + 1];
```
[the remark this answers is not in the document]
KubqoA: I was thinking, because I saw that the struct that eventually contains the text for the UI, has the attribute `char fullText[200];`, which I guess means that eventually anything longer than 200 characters won't be shown.
However, theoretically such a string is possible here with some crazy nesting, it shouldn't happen, because the native scripts shouldn't be so complex, but how should I deal with this?
The length of this array should probably be also 200, but I should probably check for eventual overflows right?

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,520 @@
...
+
+enum {
+	DISPLAY_UI_STEP_COMPLEX_SCRIPT_FINISHED = 100,
+	DISPLAY_UI_STEP_POSITION = 200,
+	DISPLAY_UI_STEP_SCRIPT_CONTENT,
+	DISPLAY_UI_STEP_RESPOND,
+	DISPLAY_UI_STEP_INVALID
+};
+
+static void deriveScriptHash_display_ui_runStep()
```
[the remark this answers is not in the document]
KubqoA: I think this is solved by removing the complex script finish call. Am I right?

src/nativeScriptHashBuilder.c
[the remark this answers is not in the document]
janmazak: Not necessarily always better, but without semicolon, you define an _expression_, not a _statement_. So you can use it in a wider variety of contexts, e.g. in an if condition. Also, imagine using a macro having a semicolon in such a way that you add an additional semicolon within an if statement without {}.

See
https://wiki.sei.cmu.edu/confluence/display/c/PRE11-C.+Do+not+conclude+macro+definitions+with+a+semicolon
for some more examples.

src/nativeScriptHashBuilder.c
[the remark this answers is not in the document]
janmazak: Generally, it's preferable to deal with enum values via a switch instead of a series of if/else statements: you can cleanly group several cases together, it is easy to modify (whether when the enum itself is modified or case handling is being modified) etc. So do not shy away from a switch even if there are just two values (usually, you also need `default` since the variable holding an enum value is not bound by its type to really only contain valid enum values).

What I wanted to say is that a function named `advanceState` suggests that we are going to look at the _current_ state and them move on to some kind of _following_ state. Your code of the function actually computed the new state from scratch without any regard to the current state. AFAIK all the advanceState/Stage functions work in the first way, not in the second. It might be OK to do it differently here, but we must have a clear reason for that.

Right now, I have noticed that `advanceState` is actually recursive? I would strongly prefer to avoid that if possible.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,458 @@
...
+			char text[35];
+			SAFE_SPRINTF(text, "Contains %d nested scripts", ctx->complexScripts[ctx->level].remainingScripts);
+
+			ui_displayPaginatedText(
+			        HEADER,
+			        text,
+			        this_fn
+			);
+		}
+		break;
```
janmazak, line 141: The location of this `break` looks suspicious --- shouldn't it go inside {}?. Also, it is important to be consistent in using {} for cases consistently --- either not at all, or for each case in a switch. I guess we should use {} whenever the cases are longer and involve indentation, as is the case here.

src/deriveNativeScriptHash.c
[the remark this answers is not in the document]
janmazak: I think yes, there isn't much left that can be done about simplifying it without annoying side-effects.

src/deriveNativeScriptHash.c
janmazak: Has this been done? I don't see it in the current version.

src/deriveNativeScriptHash.c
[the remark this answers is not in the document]
janmazak: I think we should do the same thing everywhere when we log uint64. (You are welcome to propose improvements to the existing code, e.g. adding a variable name to TRACE_UINT64.)

One purpose of tracing is that you can compare the values from usbtool to those logged in javascript (I use plain `console.log` since I don't know anything better...). Does your macro makes the comparison straightforward? If not, I would be wary of using it as a permanent solution.

src/nativeScriptHashBuilder.h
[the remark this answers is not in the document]
janmazak: Well, the header file does not need any cbor, just the implementation file does. Using "hash.h" would be sufficient (see `txHashBuilder.h`). It somehow caught my attention that the hash builder is exposing cbor stuff (AFAIK it is an internal concept for hash builders).

src/deriveNativeScriptHash.c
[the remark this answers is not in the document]
janmazak: The overflows are, at present, caught by
```
	ASSERT(body_len < SIZEOF(ctx->fullText));
```
in `ui_displayPaginatedText`. Your code should not directly deal with the UI buffer length, but it would make sense if `uiHelpers.h` offered a function that would check header and/or second line lengths and throw ERR_DATA_TOO_LARGE if it did not fit. It is definitely better to propagate a particular error code than letting Ledger just freeze on some assertion.

src/deriveNativeScriptHash.c
KubqoA: Sorry forgot to push the commit.

KubqoA: Extracted to a different function, because the level is needed later.

src/deriveNativeScriptHash.h
KubqoA: It also makes it easier to get the header, because for that I use the array called `ui_native_script_header` which translates this enum to a header for each of the script types, so if it's possible I would keep this, because it simplifies life for me in two separate places.

src/deriveNativeScriptHash.c
KubqoA: @janmazak so I assume by [this message in Slack from Gabo](https://vacuumlabs.slack.com/archives/CETDV69MF/p1628162445019000?thread_ts=1627911175.004900&cid=CETDV69MF) that I should check/make this change across the Ledger codebase, right?

src/nativeScriptHashBuilder.c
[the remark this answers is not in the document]
KubqoA: I think there is a very clear reason to do it differently here. The state is determined only by the current count of remaining scripts for the given level. Basically the state always expects another script definition, be it the beginning of a complex script or a simple script. In my opinion there's nothing to be done to distinguish between expecting a complex vs a simple script as we don't know beforehand, so it is logical that both are represented by a single enum value (`NATIVE_SCRIPT_HASH_BUILDER_SCRIPT`).
The only state changes happens only when we received everything. That's when the count of remaining scripts for the 0th level is 0, then the state changes to second enum value (`NATIVE_SCRIPT_HASH_BUILDER_FINISHED`).
I don't see any logical way to incorporate other enum values, and thus no logical way to make this `advanceState` method work as the others in the codebase do.
Also since this method is only ever called, or should ever only be called, when the state is `NATIVE_SCRIPT_HASH_BUILDER_SCRIPT` I really don't see the point of having a switch statement here. If you were to argue that the default case of the switch could throw an error like invalid state then I think that's handled by the assertion at the top of the function.
```c
static inline void advanceState(native_script_hash_builder_t* builder)
{
	// advance state should be only called when current state processed a script
	ASSERT(builder->state == NATIVE_SCRIPT_HASH_BUILDER_SCRIPT);
```
So overall, maybe the "solution" is to name this function differently or maybe I can just add a comment that the state value is derived from the count of remaining scripts? (The second option is sufficient IMHO)

src/deriveNativeScriptHash.c
janmazak: Yes.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,463 @@
...
+	if (ctx->level > 0 && ctx->complexScripts[ctx->level].remainingScripts == 0) {
+		complexScriptFinished();
+	}
+}
+
+// UI
+
+const char ui_native_script_header[7][27] = {"Script - key path", "Script - key", "Script - ALL", "Script - ANY", "Script - N of K", "Script - invalid before", "Script - invalid hereafter"};
+
+#define ASSERT_UI_SCRIPT_TYPE_SANITY() ASSERT(ctx->ui_scriptType >= UI_SCRIPT_PUBKEY_PATH && ctx->ui_scriptType <= UI_SCRIPT_INVALID_HEREAFTER);
```
janmazak, line 51:
```suggestion
#define ASSERT_UI_SCRIPT_TYPE_SANITY() ASSERT(ctx->ui_scriptType >= UI_SCRIPT_PUBKEY_PATH && ctx->ui_scriptType <= UI_SCRIPT_INVALID_HEREAFTER)
```

src/nativeScriptHashBuilder.c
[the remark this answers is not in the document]
janmazak: An short explanatory comment would be sufficient. It is especially important to highlight the recursive nature of the function (I agree we want the recursion to walk back the levels), and what the recursion does (e.g. when it ends).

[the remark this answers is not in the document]
janmazak: I'd also suggest to split the if statement; it would markedly improve readability (at least for me), also some minor changes to comments:
```
	// if the current number of remaining scripts is bigger than 0 we still
	// expect scripts
	if (builder->remainingScripts[builder->level] > 0) {
		builder->state = NATIVE_SCRIPT_HASH_BUILDER_SCRIPT;
		return;
	}

	if (builder->level > 0) {
		// we are on some level bigger than 0 and expect the end of a complex script
		builder->level--;

		ASSERT(builder->remainingScripts[builder->level] > 0);
		builder->remainingScripts[builder->level]--;

		advanceState(builder);
	} else {
		// we know the whole native script is finished
		builder->state = NATIVE_SCRIPT_HASH_BUILDER_FINISHED;
	}
```

src/nativeScriptHashBuilder.h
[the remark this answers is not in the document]
KubqoA: It needs also `cardano.h`, because it uses the `MAX_SCRIPT_DEPTH` constant.

src/deriveNativeScriptHash.c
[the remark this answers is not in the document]
KubqoA: It probably doesn't unfortunately, it displays the uint64 in it's hex form, and if I am correctly aware the `console.log` displays is as a normal number, so I will probably switch to the already existing solution

src/deriveNativeScriptHash.c
KubqoA: I tried to do it this way when I was first implementing it, but for some reason it didn't compile.

src/nativeScriptHashBuilder.c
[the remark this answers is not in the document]
KubqoA: I opted for a slightly different approach, I tried to make this function as simple as possible and added two more,
`simpleScriptFinished` and `complexScriptFinished`, analogous to the ones in the `deriveNativeScriptHash.c` file, which handle setting the correct `level` and `remainingScripts`.

src/deriveNativeScriptHash.c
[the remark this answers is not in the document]
KubqoA: I tried to implement the solution you proposed :)

src/deriveNativeScriptHash.c
KubqoA: Created an issue for this: https://github.com/vacuumlabs/ledger-app-cardano-shelley/issues/90

src/deriveNativeScriptHash.c
KubqoA: Created the issue https://github.com/vacuumlabs/ledger-app-cardano-shelley/issues/92

src/deriveNativeScriptHash.c
KubqoA: if it doesn't matter if I parse the other argument before I know the first one is valid I will fix this. Also I think there are other instances where I do something similar so I will also fix those.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,520 @@
...
+
+	UI_STEP_END(DISPLAY_UI_STEP_INVALID);
+}
+#undef HEADER
+#undef ASSERT_UI_SCRIPT_TYPE_SANITY
+
+#define UI_DISPLAY_SCRIPT(UI_TYPE)\
+	ctx->ui_scriptType = UI_TYPE;\
+	ctx->ui_step = DISPLAY_UI_STEP_POSITION;\
+	deriveScriptHash_display_ui_runStep();
```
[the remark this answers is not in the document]
KubqoA: If I am reading this correctly: https://wiki.sei.cmu.edu/confluence/display/c/PRE10-C.+Wrap+multistatement+macros+in+a+do-while+loop, then wrapping it into the `do ... while(0)` notation is mainly useful where I need to make it into one statement, so e.g. a boolean statement in an if, right?
So here it is completely okay to just use curly braces?

src/deriveNativeScriptHash.c
```diff
@@ -157,9 +157,9 @@ static void deriveScriptHash_display_ui_runStep()
 			break;
 		}
 		case UI_SCRIPT_N_OF_K: {
-			// max possible length 67: "Reuqired signatures: x. Contains x nested scripts"
-			// where x is 2^32-1
-			char text[68];
+			// max possible length 67: "Requires n out of k signatures. Contains k nested scripts."
```
janmazak, line 160: 67 is outdated?

src/uiHelpers.h
```diff
@@ -130,6 +129,9 @@ void uiCallback_reject(ui_callback_t* cb);
 void assert_uiPaginatedText_magic();
 void assert_uiPrompt_magic();
 
+void validate_uiPaginatedText_canFitStringIntoHeader(const char* str);
+void validate_uiPaginatedText_canFitStringIntoFullText(const char* str);
```
janmazak, line 133: I'd expect these would be used within `ui_displayPaginatedText` (is it true that everything we display is displayed through this function?). Also, I am slightly in favor of having these as boolean functions (so that anyone can do with their results whatever he wants, e.g. use them in an assertion or validation).

src/nativeScriptHashBuilder.c
```diff
@@ -0,0 +1,229 @@
...
+}
+
+static inline bool isComplexScriptFinished(native_script_hash_builder_t* builder)
+{
+	return builder->level > 0 && builder->remainingScripts[builder->level] == 0;
+}
+
+static void complexScriptFinished(native_script_hash_builder_t* builder)
+{
+	while(isComplexScriptFinished(builder)) {
```
janmazak, line 61:
```suggestion
	while (isComplexScriptFinished(builder)) {
```
Please check for missing spaces in other similar places (e.g. `if(`).

And `make format` needs to be run.

src/deriveNativeScriptHash.c
janmazak:
```suggestion
	TRACE_BUFFER(view->begin, VIEW_REMAINING_TO_TUPLE_BUF_SIZE(view));
```
This compiles for me.

src/deriveNativeScriptHash.c
janmazak: Hm. But it increases code size, which is annoying in the context of Ledger. Let's move them before the switch and add a comment to 
```
	uint8_t nativeScriptType = parse_u1be(view);
	TRACE("native complex script type = %u", nativeScriptType);
```
that the value is validated below.

janmazak (commented):
Just the few tiny changes and we are done.

src/deriveNativeScriptHash.c
KubqoA: But this is not correct if I am not mistaken.
```c
// Note(ppershing): these macros expand to two arguments!
#define VIEW_REMAINING_TO_TUPLE_BUF_SIZE(view) (view)->ptr, view_remainingSize(view)
#define VIEW_PROCESSED_TO_TUPLE_BUF_SIZE(view) (view)->begin, view_processedSize(view)
```
the `TRACE_BUFFER` macro expects two arguments, the pointer to the buffer and the size. So I think the correct usage should be:
```c
TRACE_BUFFER(VIEW_REMAINING_TO_TUPLE_BUF_SIZE(view));
```
However, that doesn't compile for me. Because this code is repeated I suggest we could create a macro `TRACE_REMAINING_VIEW` or similar.

Or I can move the tracing to the `deriveNativeScriptHash_handleAPDU` method, but I didn't do it at first, because of the complex native script end call, which is now deprecated. So I will probably move the tracing to this top level function and then we don't have to repeat it.

src/nativeScriptHashBuilder.c
KubqoA: If possible when I will be cleaning up the commit history I will run `make format` for each of the commits and resolve this issue then, if it's okay with you.

janmazak: Yes, I'm fine with that.

src/deriveNativeScriptHash.c
janmazak: Interesting. It does not compile because the outer macro unrolls first, and it has just a single argument at the time. Well, macros were never all-powerful :)

I prefer the current approach with tracing in `deriveNativeScriptHash_handleAPDU`.

src/nativeScriptHashBuilder.c
KubqoA: This was actually not caught and fixed by `make format`

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,464 @@
...
+
+static ins_derive_native_script_hash_context_t* ctx = &(instructionState.deriveNativeScriptHashContext);
+
+// Helper functions
+
+#define TRACE_WITH_CTX(message, ...) TRACE(message "level = %u, remaining scripts = %u", ##__VA_ARGS__, ctx->level, ctx->complexScripts[ctx->level].remainingScripts)
+#define SAFE_SPRINTF(str, format, ...) {\
+		explicit_bzero(str, SIZEOF(str));\
+		snprintf(str, SIZEOF(str), format, __VA_ARGS__);\
+		ASSERT(strlen(str) + 1 < SIZEOF(str));\
```
mkv-vcm, line 14: This is only needed if you don't trust snprintf, but then what is the point of snprintf?

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,464 @@
...
+
+static uint8_t _getScriptLevelForPosition()
+{
+	// For complex scripts we reduce the current level by 1
+	// Because they already have the level increased by 1
+	int levelOffset = ctx->ui_scriptType == UI_SCRIPT_ALL
+	                  || ctx->ui_scriptType == UI_SCRIPT_ANY
+	                  || ctx->ui_scriptType == UI_SCRIPT_N_OF_K
+	                  ? -1 : 0;
+	ASSERT(ctx->level + levelOffset >= 0);
```
mkv-vcm, line 70: I'm a bit uncomfortable with relying on the implicit conversion rules here. uint8_t + int will be int as far as I know, but it seems a bit risky that it might become an unsigned type by some accident or just not paying attention. And then the comparison with 0 is worthless.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,464 @@
...
+	TRACE();
+
+	// For level n the longest possible position is: "x.x.x...x"
+	// where the x is 2^32-1 (length 10) and there are up to level xs
+	// + the leading "Position: "
+	char position[10 + 11 * (MAX_SCRIPT_DEPTH - 1) + 1];
+	int offset = snprintf(position, 10 + 1, "Position: ");
+	for (int i = 1; i <= level; i++) {
+		snprintf(&position[offset], 11 + 1, "%u.", ctx->complexScripts[i].totalScripts - ctx->complexScripts[i].remainingScripts + 1);
+		// snprintf returns 0, https://github.com/LedgerHQ/nanos-secure-sdk/issues/28
```
mkv-vcm, line 86: In that case, then offset's initial value will also be 0, so "Position: " would be overwritten.

src/deriveNativeScriptHash.h
```diff
@@ -0,0 +1,51 @@
...
+
+	uint8_t scriptHashBuffer[SCRIPT_HASH_LENGTH];
+	native_script_hash_builder_t hashBuilder;
+
+	native_script_content_t scriptContent;
+
+	// UI information
+	int ui_step;
+	ui_native_script_type ui_scriptType;
+} ins_derive_native_script_hash_context_t;
```
mkv-vcm, line 49: Not that it's a problem necessary, but padding rules make this an inefficient layout.

KubqoA: Hm, what would be a more efficient layout, or more precisely what's making this layout inefficient?

src/deriveNativeScriptHash.c
KubqoA: For some reasons the `snprintf` function works like a black magic. With no formatting and snprinting a simple string, the return value of `snprintf` is correct. But when you include some formatting like `"%u"`, then it returns a 0.

KubqoA: I can update the comment to reflect this.

mkv-vcm: That would be helpful, because the linked issue doesn't make this obvious. Question remains if we can rely on this snprintf implementation for even not formatted text. Maybe an ASSERT to make sure offset starts at strlen("Position: ")?

src/deriveNativeScriptHash.h
mkv-vcm: https://developers.ledger.com/docs/nano-app/memory-alignment/
It'll pad out single bytes before larger (like here, 32 bit) types. If we fit into the memory that's fine, I doubt this is the largest struct in the union.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,464 @@
...
+	TRACE_WITH_CTX("simple script finished, ");
+
+	if (isComplexScriptFinished()) {
+		complexScriptFinished();
+	}
+}
+
+// UI
+
+const char ui_native_script_header[7][27] = {"Script - key path", "Script - key", "Script - ALL", "Script - ANY", "Script - N of K", "Script - invalid before", "Script - invalid hereafter"};
```
mkv-vcm, line 57: This array should hold char*s, not whole char arrays, we can spare some memory with that if the average size of the strings is less than 23 (since the pointers themselves take up 4 bytes).

src/deriveNativeScriptHash.h
janmazak: I don't think we are concerned with memory here. (We run pretty close to limits for stuff with several layers of buffers, e.g. when we are deep within tx signing, and need to derive keys and encode addresses into bech32.)

src/deriveNativeScriptHash.c
janmazak: I'd ignore the return value of snprintf, it is not reliable, rather we should always use `strlen()` to find out what it did.

Also, the value `ctx->complexScripts[i].totalScripts - ctx->complexScripts[i].remainingScripts + 1` perhaps deserves to be stored in an aptly named variable.

And maybe it makes sense to somehow name the constants (what are 10, 11 etc.)?

Ah, and one more: before using `complexScripts[i]`, we should assert the index i is not out of bounds (the condition `i <= level` looks suspicious). Consider using size_t for the for loop variable as we do elsewhere (though maybe not with 100% consistency?).

And perhaps `position` is not a suitable name for a char array, what about something like `(script)PositionDescription`?

src/deriveNativeScriptHash.c
KubqoA: Fair point, I removed it :)

src/deriveNativeScriptHash.c
KubqoA: I tried to come up with a better solution:
```c
	uint8_t levelOffset = ctx->ui_scriptType == UI_SCRIPT_ALL
	                   || ctx->ui_scriptType == UI_SCRIPT_ANY
	                   || ctx->ui_scriptType == UI_SCRIPT_N_OF_K
	                   ? 1 : 0;
	ASSERT(levelOffset == 0 || ctx->level > 0);
	return ctx->level - levelOffset;
```
I switched the type to unsigned int, and the assertion requires the offset to be either 0, or the level to be bigger than 0, which for me seems reasonable. What do you think?

src/deriveNativeScriptHash.c
KubqoA: I tried to address all your points.

src/deriveNativeScriptHash.c
KubqoA: I tried to change it, and it compiles, but Ledger crashes when working with these values. I am not sure why this happens, but now I remember that using fixed width here solved the issue for me. Maybe you will be wiser as to what is happening.

src/cardano.h
```diff
@@ -135,5 +136,18 @@ typedef struct {
 	uint8_t dnsName[DNS_NAME_SIZE_MAX];
 } pool_relay_t;
 
+// ==============================  NATIVE SCRIPTS  ==============================
+
+// depth of n means it can handle up to n-1 levels of nesting
+static const size_t MAX_SCRIPT_DEPTH = 11;
```
janmazak, line 142: Do we have a STATIC_ASSERT making sure that levels (i.e. `MAX_SCRIPT_DEPTH`) fits into `uint8_t`? It would fit nicely somewhere into the following piece of code:
```
	// check if we can increase the level without breaking the MAX_SCRIPT_DEPTH constraint
	VALIDATE(ctx->level + 1 < MAX_SCRIPT_DEPTH, ERR_INVALID_DATA);
	ctx->level++;

```
[While this is very unlikely to ever create any problems, we should practice not neglecting such considerations.]

src/deriveNativeScriptHash.c
janmazak: I'm OK with this.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,469 @@
...
+	ASSERT_UI_SCRIPT_TYPE_SANITY();
+	TRACE();
+
+	const size_t POSITION_PREFIX_LENGTH = 11;
+	const size_t MAX_LEVEL_LENGTH = 12;
+
+	// For level n the longest possible position is: "x.x.x...x"
+	// where the x is 2^32-1 (length 10) and there are up to level xs
+	// + the leading "Position: "
+	char positionDescription[10 + 11 * (MAX_SCRIPT_DEPTH - 1) + 1];
```
janmazak, line 80: I would prefer to also apply `explicit_bzero` here.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,469 @@
...
+
+	for (size_t i = 1; i <= level; i++) {
+		ASSERT(i < MAX_SCRIPT_DEPTH);
+		uint32_t position = ctx->complexScripts[i].totalScripts - ctx->complexScripts[i].remainingScripts + 1;
+		snprintf(&positionDescription[offset], MAX_LEVEL_LENGTH, "%u.", position);
+		offset += strlen(&positionDescription[offset]);
+	}
+
+	// remove any trailing '.'
+	positionDescription[offset - 1] = '\0';
```
janmazak, line 95: Let's assert `offset > 0`.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,469 @@
...
+	// snprintf returns 0, https://github.com/LedgerHQ/nanos-secure-sdk/issues/28
+	// so we need to check the number of written characters by `strlen`
+	snprintf(positionDescription, POSITION_PREFIX_LENGTH, "Position: ");
+	int offset = strlen(positionDescription);
+
+	for (size_t i = 1; i <= level; i++) {
+		ASSERT(i < MAX_SCRIPT_DEPTH);
+		uint32_t position = ctx->complexScripts[i].totalScripts - ctx->complexScripts[i].remainingScripts + 1;
+		snprintf(&positionDescription[offset], MAX_LEVEL_LENGTH, "%u.", position);
+		offset += strlen(&positionDescription[offset]);
```
janmazak, line 91:
```suggestion
		offset += strlen(positionDescription + offset);
```

src/deriveNativeScriptHash.c
janmazak: Also, it is not obvious that the last char actually is a dot (i.e. why the for cycle is run at least once) --- it depends on the given parameter `level`, and so this function should assert it is given a meaningful value for it.

src/deriveNativeScriptHash.c
```diff
@@ -0,0 +1,469 @@
...
+
+	// snprintf returns 0, https://github.com/LedgerHQ/nanos-secure-sdk/issues/28
+	// so we need to check the number of written characters by `strlen`
+	snprintf(positionDescription, POSITION_PREFIX_LENGTH, "Position: ");
+	int offset = strlen(positionDescription);
+
+	for (size_t i = 1; i <= level; i++) {
+		ASSERT(i < MAX_SCRIPT_DEPTH);
+		uint32_t position = ctx->complexScripts[i].totalScripts - ctx->complexScripts[i].remainingScripts + 1;
+		snprintf(&positionDescription[offset], MAX_LEVEL_LENGTH, "%u.", position);
```
janmazak, line 90: I'd prefer if we used remaining buffer size in `snprintf` (everywhere), and asserted no problem happened after each snprintf (e.g. using `strlen`). Otherwise, a simple change of the constants POSITION_PREFIX_LENGTH etc. could lead to a buffer overflow. (See `str_formatAdaAmount` for an example.)

src/deriveNativeScriptHash.c
janmazak: Static strings are stored differently than code, and that leads to subtle problems with pointers. 
Take a look at
https://developers.ledger.com/docs/nano-app/position-independent-code/
There are macros to deal with it (namely, `PIC` and our own `PTR_PIC`, see `run_base58_test` for an example).

src/cardano.h
KubqoA: Wouldn't it be better to just then use `uint8_t` as the type of the `MAX_SCRIPT_DEPTH`?

src/deriveNativeScriptHash.c
KubqoA: Is it any different than using:
```c
char positionDescription[10 + 11 * (MAX_SCRIPT_DEPTH - 1) + 1] = {0};
```
because if I remember correctly this should set all the chars in the array to be zero bytes.

src/cardano.h
janmazak: Yes, true.

src/deriveNativeScriptHash.c
janmazak: It is true according to some standard (C99?). But for some reason, we do not rely on that and use `explicit_bzero` elsewhere. I am not sure how the app is actually compiled and built by Ledger, they might circumvent our compiler options, so I would stick with what `explicit_bzero`.

mkv-vcm: After snprintf, I wouldn't feel safe relying on the standard implementation too much.

src/deriveNativeScriptHash.c
KubqoA: I added an assertion whether the level is bigger than 0.

src/deriveNativeScriptHash.c
KubqoA: Added the `explicit_bzero`.

src/deriveNativeScriptHash.c
KubqoA: https://github.com/vacuumlabs/ledger-app-cardano-shelley/issues/96

PR #88, opened 2021-08-05
multisig

mkv-vcm:


src/cbor.c
```diff
@@ -112,6 +126,16 @@ size_t cbor_writeToken(uint8_t type, uint64_t value, uint8_t* buffer, size_t buf
 
 	// Check sanity
 	switch (type) {
+	case CBOR_TYPE_NEGATIVE: {
+		int64_t negativeValue;
+		// reinterpret an actually negative value hidden in an unsigned in the safe way
+		memcpy(&negativeValue, &value, SIZEOF(value));
```
mkv-vcm, line 132: > I guess we want some STATIC_ASSERT before this, e.g. SIZEOF(value) == SIZEOF(negativeValue).

I guess we could trust that 64_t types are the same size, but someone can switch it under us, so yeah, I'll add it. Doesn't cost us application size.

src/cbor.c
```diff
@@ -45,49 +46,62 @@ cbor_token_t cbor_parseToken(const uint8_t* buf, size_t size)
...
+		if (result.value < limit_min) {
+			// This isn't canonical CBOR
+			THROW(ERR_UNEXPECTED_TOKEN);
+		}
 	}
 
-	if (result.value < limit_min) {
-		// This isn't canonical CBOR
-		THROW(ERR_UNEXPECTED_TOKEN);
+	if (CBOR_TYPE_NEGATIVE == result.type) {
```
mkv-vcm, line 94: > I think we don't need to parse CBOR_TYPE_NEGATIVE. It is OK to have it here, but it should probably be just commented out (to avoid increasing app size) if we ever needed it in the future. (AFAIK we are only parsing CBOR for some legacy Byron stuff, like extracting protocol magic, or something like that.) But the size difference is perhaps negligible, so I guess I am OK with having it here (especially if we can't easily comment it out in a single place, I've not really checked).

I'm not certain _why_ we don't want to parse negatives here, I'm not clear on what type of CBORs the app will parse.

mkv-vcm:
> In key derivation path validations, notice that role cannot be 1 for multisig keys.
> https://cips.cardano.org/cips/cip1854/
> https://cips.cardano.org/cips/cip1852/
> We need to make sure we are compliant with both of these.

src/cbor.c
janmazak: CBOR parsing is actually legacy stuff from Byron. All the new Shelley stuff only serializes into CBOR, but with Byron, some parsing is unavoidable (you need to extract protocol magic from address or something like that; notice that the only place where CBOR is parsed is the call `view_readToken` in addressUtilsByron). It is quite possible that we might cut down the parsing even more than it is, no one though about it much because no one looked at legacy Byron address format in detail. Perhaps you can create a github issue for this.

mkv-vcm: Then I understand, CBOR_NEGATIVE is only used for minting, which is not a Byron issue.
A github issue for ledger? Or an in house one?

janmazak: Just an in-house one, in this repository, so that at a more opportune time we might take a look if we were interested in decreasing app size.

src/signTx.c
```diff
@@ -1395,11 +1393,27 @@ static void signTx_handleWithdrawal_ui_runStep()
...
+			};
+			constructRewardAddressFromHash(
+				ctx->commonTxData.networkId,
+				REWARD_HASH_SOURCE_SCRIPT,
+				txBodyCtx->stageData.withdrawal.stakeCredential.scriptHash,
+				SIZEOF(txBodyCtx->stageData.withdrawal.stakeCredential.scriptHash),
+				rewardAccount.hashBuffer,
+				SIZEOF(rewardAccount.hashBuffer)
+			);
+			ui_displayRewardAccountScreen(&rewardAccount, ctx->commonTxData.networkId, this_fn);
```
janmazak, line 1414: Ideally, we want to end the `UI_STEP` with a single call to `ui_display...`, not having a ui call in every if branch.

src/signTx.c
```diff
@@ -1457,12 +1489,9 @@ static void signTx_handleWithdrawalAPDU(uint8_t p2, uint8_t* wireDataBuffer, siz
 		read_view_t view = make_read_view(wireDataBuffer, wireDataBuffer + wireDataSize);
 		VALIDATE(view_remainingSize(&view) >= 8, ERR_INVALID_DATA);
 		txBodyCtx->stageData.withdrawal.amount = parse_u8be(&view);
-		// the rest is path
+		// the rest is the identifier, either path or scripthash
```
janmazak, line 1492: identifier -> stake credential?
Or I would remove the comment, the next line is self-explanatory.

src/signTx.h
```diff
@@ -63,14 +63,17 @@ typedef enum {
 } stake_credential_type_t; // TODO rename to stake credential type
```
janmazak, line 63: No need for the TODO anymore.

src/securityPolicy.c
```diff
@@ -634,51 +636,81 @@ security_policy_t policyForSignTxStakePoolRegistrationConfirm(
...
-		default:
-			DENY();
-			break;
-		}
-
-		break;
-	}
+	case SIGN_TX_USECASE_MULTISIG:
+		DENY_UNLESS(multisig || mintPresent && mint);
+		return nonPoolSubPolicy(pathSpec);
```
janmazak, line 713: For multisig, everything has to be shown, `nonPoolSubPolicy` returns ALLOW sometimes.

src/securityPolicy.c
```diff
@@ -634,51 +636,81 @@ security_policy_t policyForSignTxStakePoolRegistrationConfirm(
...
 // previously declared inputs and certificates
 security_policy_t policyForSignTxWitness(
         sign_tx_usecase_t signTxUsecase,
-        const bip44_path_t* pathSpec
+        const bip44_path_t* pathSpec,
+		bool mintPresent
 )
 {
+	const bool ordinary = bip44_isOrdinarySpendingKeyPath(pathSpec);
+	const bool multisig = bip44_isMultisigSpendingKeyPath(pathSpec);
```
janmazak, line 702: I don't understand the intent here, are we ignoring staking paths?

src/bip44.c
```diff
@@ -192,6 +192,20 @@ bool bip44_isOrdinarySpendingKeyPath(const bip44_path_t* pathSpec)
...
+	       bip44_hasValidChainTypeForAddress(pathSpec) &&
+	       bip44_containsAddress(pathSpec);
+}
+
+bool bip44_isMintSpendingKeyPath(const bip44_path_t* pathSpec)
+{
+	return bip44_hasMintKeyPrefix(pathSpec) &&
+	       bip44_hasValidChainTypeForAddress(pathSpec) &&
+	       bip44_containsAddress(pathSpec);
+}
```
janmazak: This is totally wrong. See https://github.com/cardano-foundation/CIPs/blob/master/CIP-1855/CIP-1855.md for minting paths. I think we should rely on `bip44_classifyPath` in security policies (just deal with each possible path from the enum in a switch statement).

src/securityPolicy.c
```diff
@@ -634,51 +636,81 @@ security_policy_t policyForSignTxStakePoolRegistrationConfirm(
...
+		DENY();
+		break;
+	}
+}
+
 // For each transaction witness
 // Note: witnesses reveal public key of an address
 // and Ledger *does not* check whether they correspond to
 // previously declared inputs and certificates
 security_policy_t policyForSignTxWitness(
```
janmazak, line 695: I imagine that this function should only consist of a switch statement that distinguishes the usecase and then for each use case, we call a separate function that would contain a switch statement distinguishing cases for the `bip44_classifyPath` result. Then it would be straightforward and easily auditable.

mkv-vcm: That's basically what is happening right now, but we have TWO things to decide (deny/allow and show/prompt/warn). Right now I separated those two for the non-pool use cases, and I think it looks better for it. But I'm open for ideas.

src/bip44.c
mkv-vcm: I don't think in this particular case the classification would help, it would just bulk up the code. But true, I missed the memo on the 1855 paths. Fixed.

src/securityPolicy.c
mkv-vcm: We did, but the trezor test showed this, and got fixed.

src/securityPolicy.c
mkv-vcm: Done.

src/securityPolicy.c
```diff
@@ -634,51 +636,81 @@ security_policy_t policyForSignTxStakePoolRegistrationConfirm(
...
+
+	default:
+		ASSERT(false);
+	}
+
+	DENY(); // should not be reached
 }
 
 
+static inline security_policy_t nonPoolSubPolicy(const bip44_path_t* pathSpec)
```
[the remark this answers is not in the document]
mkv-vcm: It's called from the transaction use cases that are not pool registration. It's just a static function, so I think context can be used instead of an overly verbose name, but if you have a good suggestion, I'm all for it.

src/signTx.h
mkv-vcm: Done.

src/signTx.c
mkv-vcm: Removed.

src/signTx.c
mkv-vcm: Done, though it makes the variable live in a larger scope. This is the one return or multiple returns consideration just a different colour of it.

src/signTx.c
```diff
@@ -1395,11 +1393,27 @@ static void signTx_handleWithdrawal_ui_runStep()
...
-			.path = txBodyCtx->stageData.withdrawal.path
-		};
-		ui_displayRewardAccountScreen(&rewardAccount, ctx->commonTxData.networkId, this_fn);
+		if (STAKE_CREDENTIAL_KEY_PATH == txBodyCtx->stageData.withdrawal.stakeCredential.type) {
+			reward_account_t rewardAccount = {
+				.keyReferenceType = KEY_REFERENCE_PATH,
+				.path = txBodyCtx->stageData.withdrawal.stakeCredential.pathSpec
+			};
+			ui_displayRewardAccountScreen(&rewardAccount, ctx->commonTxData.networkId, this_fn);
+		} else {
```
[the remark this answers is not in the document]
mkv-vcm: Done.

src/signTx.c
```diff
@@ -1348,10 +1344,12 @@ static void signTx_handleCertificateAPDU(uint8_t p2, uint8_t* wireDataBuffer, si
 
 	case CERTIFICATE_TYPE_STAKE_POOL_RETIREMENT: {
 		security_policy_t policy = policyForSignTxCertificateStakePoolRetirement(
-		                                   ctx->commonTxData.signTxUsecase,
-		                                   &txBodyCtx->stageData.certificate.pathSpec,
-		                                   txBodyCtx->stageData.certificate.epoch
-		                           );
+										ctx->commonTxData.signTxUsecase,
```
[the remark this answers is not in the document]
mkv-vcm: Not sure what makes that "complete", but it is handled in the policy, I don't think this code should know about it.

src/signTx.c
```diff
@@ -1123,23 +1123,18 @@ static void _parsePathSpec(read_view_t* view, bip44_path_t* pathSpec)
...
+		_parsePathSpec(view, &identifier->pathSpec);
 		break;
-	}
-
 	case STAKE_CREDENTIAL_SCRIPT_HASH: {
-		STATIC_ASSERT(SIZEOF(certificateData->scriptHash) == SCRIPT_HASH_LENGTH, "bad script hash container size");
-		VALIDATE(view_remainingSize(view) >= SIZEOF(certificateData->scriptHash), ERR_INVALID_DATA);
-		view_memmove(certificateData->scriptHash, view, SIZEOF(certificateData->scriptHash));
+		STATIC_ASSERT(SIZEOF(identifier->scriptHash) == SCRIPT_HASH_LENGTH, "bad script hash container size");
+		VALIDATE(SIZEOF(identifier->scriptHash) <= view_remainingSize(view), ERR_INVALID_DATA);
```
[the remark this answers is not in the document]
mkv-vcm: Habit, protection from accidental assignments, and makes it much more clear when an assignment is intentional. We can purge these out of if agree that this is not wanted in the code, but I would do it in a separate commit.

src/signTx.c
```diff
@@ -1123,23 +1123,18 @@ static void _parsePathSpec(read_view_t* view, bip44_path_t* pathSpec)
 	BIP44_PRINTF(pathSpec);
 }
 
-static void _parseStakeCredentialType(read_view_t* view, sign_tx_certificate_data_t* certificateData)
+static void _parseStakeCredential(read_view_t* view, stake_credential_t* identifier)
```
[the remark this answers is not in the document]
mkv-vcm: Done.

src/signTx.c
```diff
@@ -1017,17 +1017,17 @@ static void signTx_handleCertificate_ui_runStep()
...
-		if (STAKE_CREDENTIAL_KEY_PATH == txBodyCtx->stageData.certificate.stakeCredentialType) {
+		if (STAKE_CREDENTIAL_KEY_PATH == txBodyCtx->stageData.certificate.stakeCredential.type) {
 			ui_displayPathScreen(
 			        "Staking key",
-			        &txBodyCtx->stageData.certificate.pathSpec,
+			        &txBodyCtx->stageData.certificate.stakeCredential.pathSpec,
 			        this_fn
 			);
 		} else {
 			ui_displayHexBufferScreen(
```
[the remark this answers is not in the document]
mkv-vcm: Done.

src/securityPolicy.c
```diff
@@ -491,7 +491,9 @@ security_policy_t policyForSignTxCertificateStakePoolRetirement(
 
 	case SIGN_TX_USECASE_ORDINARY_TX:
 	case SIGN_TX_USECASE_MULTISIG:
-		DENY_UNLESS(bip44_isValidPoolColdKeyPath(poolIdPath));
+		if (poolIdPath != NULL) {
```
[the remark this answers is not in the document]
mkv-vcm: Done.

src/addressUtilsShelley.h
```diff
@@ -94,10 +94,17 @@ __noinline_due_to_stack__
 size_t constructRewardAddressFromKeyPath(
         const bip44_path_t* path, uint8_t networkId, uint8_t* outBuffer, size_t outSize
 );
+
+typedef enum {
+	REWARD_HASH_SOURCE_KEY,
+	REWARD_HASH_SOURCE_SCRIPT,
+} reward_address_hash_source_t;
```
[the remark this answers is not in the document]
mkv-vcm: I don't really see the similarities.

src/signTx.c
janmazak: Only partially --- `return` ends the function while `ui_` does not.

src/securityPolicy.c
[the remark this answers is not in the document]
mkv-vcm:
```
	case SIGN_TX_USECASE_ORDINARY_TX: {

		switch (bip44_classifyPath(pathSpec)) {

		case PATH_WALLET_SPENDING_KEY:
		case PATH_WALLET_STAKING_KEY:
		case PATH_POOL_COLD_KEY:
			if (bip44_isPathReasonable(pathSpec)) {
				ALLOW();
			} else {
				WARN();
			}
			break;

		default:
			DENY();
			break;
		}

		break;
	}

```
This is how it looked before. The nonpool refers to the tx usecase, not the path's poolness.

src/signTx.c
```diff
@@ -1127,6 +1198,30 @@ static void _parseCertificateData(uint8_t* wireDataBuffer, size_t wireDataSize,
 	ASSERT(view_remainingSize(&view) == 0);
 }
 
+__noinline_due_to_stack__
```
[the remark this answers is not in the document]
mkv-vcm: Good observation. Done.

src/signTx.c
```diff
@@ -1189,6 +1277,7 @@ static void _addCertificateDataToTx(
 __noinline_due_to_stack__
 static void signTx_handleCertificateAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wireDataSize)
 {
+	// scripthash must be added next to paths
```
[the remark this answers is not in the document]
mkv-vcm: As far as I remember the intent of the comment, probably yes. The code now parses stake credential instead of just path.

src/signTx.c
```diff
@@ -1372,20 +1505,21 @@ static void signTx_handleWithdrawalAPDU(uint8_t p2, uint8_t* wireDataBuffer, siz
...
 
 	}
 
 	_addWithdrawalToTxHash();
 
-	security_policy_t policy = policyForSignTxWithdrawal();
+	security_policy_t policy = policyForSignTxWithdrawal(
+		ctx->commonTxData.signTxUsecase,
+		txBodyCtx->stageData.withdrawal.stakeCredential.type,
+		txBodyCtx->stageData.withdrawal.stakeCredential.type == STAKE_CREDENTIAL_KEY_PATH ?
```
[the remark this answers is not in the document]
mkv-vcm: I just passed the credential here as well, seems more akin to your previous intent.

src/securityPolicy.c
```diff
@@ -476,24 +484,27 @@ security_policy_t policyForSignTxCertificateStaking(
 		ASSERT(false);
 	}
 
-	if (stakingKeyPath) {
-		DENY_UNLESS(bip44_isOrdinaryStakingKeyPath(stakingKeyPath));
+	if (stakeCredential->type == STAKE_CREDENTIAL_KEY_PATH) {
+		DENY_UNLESS(bip44_isOrdinaryStakingKeyPath(&stakeCredential->pathSpec));
```
janmazak, line 488: I'd move `switch (certificateType)` to the beginning of the function, and then add this check to `case SIGN_TX_USECASE_ORDINARY_TX`.

src/securityPolicy.c
```diff
@@ -476,24 +484,27 @@ security_policy_t policyForSignTxCertificateStaking(
...
 
 security_policy_t policyForSignTxCertificateStakePoolRetirement(
         sign_tx_usecase_t signTxUsecase,
-        const bip44_path_t* poolIdPath,
+        const stake_credential_t* stakeCredential,
         uint64_t epoch MARK_UNUSED
 )
+// not sure about this one to be fair, the fact that we store
+// the pool id path in the stake credentials is a memory saving measure,
+// the policy wouldn't have to know about it
```
janmazak, line 501: Yeah, the policy should know nothing about that, it should only be given a path.

src/signTx.h
```diff
@@ -52,10 +57,18 @@ typedef struct {
...
+} stake_credential_t;
+
 typedef struct {
 	certificate_type_t type;
-	bip44_path_t pathSpec; // interpretation depends on type // TODO rename to keyPath?
 
+	stake_credential_t stakeCredential;
 	// only for specific types
 	uint8_t poolKeyHash[POOL_KEY_HASH_LENGTH];
 	uint64_t epoch;
```
janmazak, line 74:
```suggestion
	// certificates are heterogenous, so the fields are relevant only for certain types
	// the grouping into unions is only used to save some memory
	union {
		stake_credential_t stakeCredential;
		uint64_t epoch;
	};
	uint8_t poolKeyHash[POOL_KEY_HASH_LENGTH];
```

janmazak (commented):
`make format` should be run.

src/securityPolicy.c
```diff
@@ -565,51 +649,125 @@ security_policy_t policyForSignTxStakePoolRegistrationConfirm(
...
+	switch (bip44_classifyPath(pathSpec)) {
+	case PATH_ORDINARY_SPENDING_KEY:
+	case PATH_ORDINARY_STAKING_KEY:
+	case PATH_POOL_COLD_KEY:
+		if (bip44_isPathReasonable(pathSpec)) {
+			ALLOW();
+		} else {
+			WARN();
+		}
+		break;
```
janmazak, line 719: I'd welcome a blank line separating the more complicated cases in switch statements.

src/securityPolicy.c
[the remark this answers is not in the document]
janmazak: This `nonPoolSubPolicy` is no longer needed?

src/signTx.h
janmazak: I also think that adding pool key path to the anonymous union (or somewhere in the union structure where it makes sense) is preferable to storing it in stakeCredential; we might understand what's going on now, but in a couple of months it would look odd.

src/addressUtilsShelley.h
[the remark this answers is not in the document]
janmazak: `staking_data_source_t` is a superset of what we need here: it has values STAKING_KEY_HASH and STAKING_SCRIPT_HASH that can be used to distinguish reward address type (we only need it for address header if I understand it correctly). (We can assert that other staking_data_source_t enum values are not used in this context.)

src/securityPolicy.c
```diff
@@ -296,6 +347,9 @@ security_policy_t policyForSignTxOutputAddressParams(
 
 		break;
 	}
+	case SIGN_TX_USECASE_MULTISIG:
+		DENY();
+		break;
```
janmazak, line 352: It'd be best if all cases used {} consistently. (And I'd welcome a blank line separating them.)

mkv-vcm: If you wish so!

src/securityPolicy.c
mkv-vcm: There you go.

src/signTx.h
mkv-vcm: Grouped all the pool retirement data into one struct, in union with the stake credentials, since these two are mutually exclusive.

src/securityPolicy.c
mkv-vcm: Reworked.

src/securityPolicy.c
mkv-vcm: Good call, multisig doesn't have pathSpec.

src/securityPolicy.c
```diff
@@ -410,24 +476,44 @@ security_policy_t policyForSignTxCertificateStaking(
 		ASSERT(false);
 	}
 
-	DENY_UNLESS(bip44_isValidStakingKeyPath(stakingKeyPath));
+	switch (txSigningMode) {
+	case SIGN_TX_SIGNINGMODE_ORDINARY_TX:
+		DENY_UNLESS(stakeCredential->type == STAKE_CREDENTIAL_KEY_PATH);
+		if (stakeCredential->type == STAKE_CREDENTIAL_KEY_PATH) {
```
janmazak, line 482: The if condition is always true.

src/securityPolicy.c
janmazak: The comment should be removed, too.

src/signTx.h
```diff
@@ -33,32 +35,48 @@ typedef enum {
...
-	uint64_t epoch;
+	union {
+		stake_credential_t stakeCredential;
+		struct {
+			bip44_path_t poolIdPath;
+			uint64_t epoch;
+			uint8_t poolKeyHash[POOL_KEY_HASH_LENGTH];
+		};
+	};
+
```
janmazak, line 79: We need to store pool id for stake_delegation. Right now, it goes into poolKeyHash, so it overlaps with stakeCredential within this union.

src/signTx.h
```diff
@@ -33,32 +35,48 @@ typedef enum {
...
+	sign_tx_signingmode_t txSigningMode;
 
 	uint8_t networkId; // part of Shelley address
 	uint32_t protocolMagic; // part of Byron address
 } common_tx_data_t;
 
+typedef struct {
+	stake_credential_type_t type;
+	union {
+		bip44_path_t pathSpec; // interpretation depends on type // TODO rename to keyPath?
```
janmazak, line 63: This comment should be dealt with, though I'm not sure how at the moment.

src/signTxMint.c
```diff
@@ -24,7 +24,7 @@ static inline void advanceState()
 	switch (subctx->state) {
 
 	case STATE_MINT_TOP_LEVEL_DATA:
-		ASSERT(subctx->numAssetGroups > 0);
+		VALIDATE(subctx->numAssetGroups > 0, ERR_INVALID_DATA);
```
janmazak, line 27: The assertion should stay here, it only ensures that the assumptions under which the state machine works are satisfied. If it was zero at this point, we have a bug somewhere.

src/signTxMint.c
```diff
@@ -33,7 +33,7 @@ static inline void advanceState()
 		ASSERT(subctx->currentAssetGroup < subctx->numAssetGroups);
 
 		// we are going to receive token amounts for this group
-		ASSERT(subctx->numTokens > 0);
+		VALIDATE(subctx->numTokens > 0, ERR_INVALID_DATA);
```
janmazak, line 36: The value of `subctx->numTokens` should be validated immediately after it is received (as we do with other values). The assertion is all right here, ensures that we have not messed up something before (and we did, by forgetting the validation).

src/signTxOutput.c
```diff
@@ -70,7 +70,7 @@ static inline void advanceState()
 		ASSERT(subctx->currentAssetGroup < subctx->numAssetGroups);
 
 		// we are going to receive token amounts for this group
-		ASSERT(subctx->numTokens > 0);
+		VALIDATE(subctx->numTokens > 0, ERR_INVALID_DATA);
```
janmazak, line 73: The same as above, assertion is OK here.

mkv-vcm: See above.

src/signTxMint.c
mkv-vcm: See above.

src/signTxMint.c
mkv-vcm: True, we should validate at the point of reading it.

src/signTx.h
mkv-vcm: Not that it matters that much, but yeah, scriptHash and keyPath sound like a good duo.

src/signTx.h
mkv-vcm: True. Looking through the code, only the stakeCredential and the poolIdPath is really suitable to be in a union.

src/securityPolicy.c
mkv-vcm: True, this remnant is fixed.

src/signTx.h
janmazak: Let's name it `keyPath` then.

src/addressUtilsShelley.h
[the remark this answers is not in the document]
mkv-vcm: I really don't see the advantage of this. It's not two enums for the same purpose, one is a general enum for staking data sources, the other is only telling a function what the hash came from. At every call of this function it's explicitly given as a literal, not from a variable, which further suggests to me that this enum is appropriate.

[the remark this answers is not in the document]
janmazak: OK. Let's leave it as it is, then.

src/signTx.h
mkv-vcm: Not sure if we should rename it everywhere else as well where bip44_path_t exists.

janmazak: Not now, if at all.

src/signTx.c
```diff
@@ -459,45 +490,10 @@ static void signTx_handleInitAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
...
-			// to a given utxo, withdrawal or certificate.
-
-			size_t maxNumWitnesses = 0;
-			switch (ctx->commonTxData.signTxUsecase) {
-			case SIGN_TX_USECASE_POOL_REGISTRATION_OWNER:
-				maxNumWitnesses = 1;
-				break;
-
-			case SIGN_TX_USECASE_POOL_REGISTRATION_OPERATOR:
-				ASSERT(ctx->numCertificates == 1);
```
mkv-vcm, line 476: This check is also validated above, so it was superfluous.

PR #98, opened 2021-08-27
Multisig final version

janmazak:
Should work with multisig-final branch of ledgerjs.

### DIFF
--- a/src/uiHelpers.c
+++ b/src/uiHelpers.c
@@ -225,14 +225,10 @@ void respond_with_user_reject()
 
 bool uiPaginatedText_canFitStringIntoHeader(const char *str)
 {
-	paginatedTextState_t* ctx = paginatedTextState;
-
-	return strlen(str) < SIZEOF(ctx->header);
+	return strlen(str) < SIZEOF(paginatedTextState->header);
 }
 
 bool uiPaginatedText_canFitStringIntoFullText(const char *str)
 {
-	paginatedTextState_t* ctx = paginatedTextState;
-
-	return strlen(str) < SIZEOF(ctx->fullText);
+	return strlen(str) < SIZEOF(paginatedTextState->fullText);
 }


src/securityPolicy.c
```diff
@@ -146,53 +145,111 @@ security_policy_t policyForReturnDeriveAddress(const addressParams_t* addressPar
...
+	PROMPT();
 }
 
 // Derive address and show it to the user
 security_policy_t policyForShowDeriveAddress(const addressParams_t* addressParams)
 {
 	DENY_UNLESS(isValidAddressParams(addressParams));
 
-	switch (bip44_classifyPath(&addressParams->spendingKeyPath)) {
+	switch (addressParams->type) {
```
gabrielKerekes, line 201: Do `policyForReturnDeriveAddress` and `policyForShowDeriveAddress` need to be separated as they are the same except for the last `SHOW()`/`PROMPT()`? Wouldn't it make sense to extract the common part into for example `policyForDeriveAddress`?

src/securityPolicy.c
```diff
@@ -565,53 +640,99 @@ security_policy_t policyForSignTxStakePoolRegistrationConfirm(
...
-		default:
-			DENY();
-			break;
+	default:
+		DENY();
+		break;
+	}
+}
+
+static inline security_policy_t multisigWitnessPolicy(const bip44_path_t* pathSpec, bool mintPresent)
```
gabrielKerekes, line 691: This should perhaps be called `scriptWitnessPolicy`? Since it's only call under `SIGN_TX_SIGNINGMODE_SCRIPT_TX` and "multisig" is now only used elsewhere in connection with the 1854 paths.

src/securityPolicy.c
```diff
@@ -565,53 +640,99 @@ security_policy_t policyForSignTxStakePoolRegistrationConfirm(
...
+{
+	switch (txSigningMode) {
+	case SIGN_TX_SIGNINGMODE_ORDINARY_TX:
+		return ordinaryWitnessPolicy(pathSpec, mintPresent);
 
-	case SIGN_TX_USECASE_POOL_REGISTRATION_OWNER: {
+	case SIGN_TX_SIGNINGMODE_SCRIPT_TX:
+		return multisigWitnessPolicy(pathSpec, mintPresent);
 
+	case SIGN_TX_SIGNINGMODE_POOL_REGISTRATION_OWNER: {
```
gabrielKerekes, line 731: Perhaps the two remaining case statements could also be extracted out into functions?

gabrielKerekes (approved):
Untested approve. The policies look good to me. I just found some refactoring opportunities perhaps. When you confirm this is final I will test it manually to confirm everything's good.

src/securityPolicy.c
janmazak: Yes, saved 190 bytes of code :)

src/bip44.c
```diff
@@ -247,46 +313,94 @@ size_t bip44_printToStr(const bip44_path_t* pathSpec, char* out, size_t outSize)
...
+		       PATH_ORDINARY_ACCOUNT :
+		       PATH_INVALID;
+	}
+	case 5: {
+		const uint8_t chainType = bip44_getChainTypeValue(pathSpec);
+		switch (chainType) {
+
+		case CARDANO_CHAIN_INTERNAL:
+		case CARDANO_CHAIN_EXTERNAL:
+			// we do not exclude hardened address index
```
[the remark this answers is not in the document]
gabrielKerekes: I see. But for some reason the comment is still confusing. It's not obvious to me why we should exclude it. Maybe the comment could say `We don't need to check whether the address index is hardened or not as both are allowed. However, a warning is shown for addresses with hardened address index in certain places as it is unusual and should not be used.` Or something like that. It should explain why we don't exclude them. However if you think it's OK, I'm good with it too 😉

src/securityPolicy.c
gabrielKerekes: Hah, nice 👍

src/bip44.c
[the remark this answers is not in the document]
janmazak: I have replaced the comment with 
			// We do not exclude hardened address index for legacy reasons;
			// such indices have been allowed since Byron and no one really knows if they are in use,
			// so we don't want to make users' funds on such addresses unavailable.
			// But such addresses are given a warning
			// and are never hidden from users (see bip44_isPathReasonable).
which is much more clear.

Also, hardened address index need not to be allowed for multisig since the "legacy reasons" argument does not apply, we never supported such paths.

src/securityPolicy.c
```diff
@@ -565,86 +656,124 @@ security_policy_t policyForSignTxStakePoolRegistrationConfirm(
...
+	DENY(); // should not be reached
+}
+
+static inline security_policy_t _ordinaryWitnessPolicy(const bip44_path_t* path, bool mintPresent)
 {
-	switch (signTxUsecase) {
+	switch (bip44_classifyPath(path)) {
+	case PATH_ORDINARY_SPENDING_KEY:
+	case PATH_ORDINARY_STAKING_KEY:
+	case PATH_POOL_COLD_KEY:
```
refi93, line 688: should we really allow pool cold keys in the "ordinary" case? Isn't that reserved only for the case when signing a pool registration as an owner?

janmazak: They are also used for pool retirement certificates that are only included in "ordinary" transactions. But there is no reason to allow it for script tx, so I've removed it from there.

PR #100, opened 2021-08-30
refactor int64 parsing

janmazak:


mkv-vcm (approved):
Good, makes it more clear.

PR #101, opened 2021-08-30
Parsing

janmazak:
Reviewers: it must be checked that there are buffer size validations in place everywhere, and that in every case we validate that there are no excessive data in wireBuffer, which is usually done by
```
	VALIDATE(view_remainingSize(view) == 0, ERR_INVALID_DATA);
```

### DIFF
--- a/src/cardano.c
+++ b/src/cardano.c
@@ -11,7 +11,7 @@ void rewardAccountToBuffer(
 	switch (rewardAccount->keyReferenceType) {
 
 	case KEY_REFERENCE_HASH: {
-		ASSERT(SIZEOF(rewardAccount->hashBuffer) == REWARD_ACCOUNT_SIZE);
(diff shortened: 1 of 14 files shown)


src/cardano.c
```diff
@@ -11,7 +11,7 @@ void rewardAccountToBuffer(
 	switch (rewardAccount->keyReferenceType) {
 
 	case KEY_REFERENCE_HASH: {
-		STATIC_ASSERT(SIZEOF(rewardAccount->hashBuffer) == REWARD_ACCOUNT_SIZE, "wrong reward account size");
+		ASSERT(SIZEOF(rewardAccount->hashBuffer) == REWARD_ACCOUNT_SIZE);
```
mkv-vcm, line 14: Why is an ASSERT better in this case? These are all compile time information.

mkv-vcm (approved):

PR #102, opened 2021-08-30
remove bech32prefix TODO

### LINKED ISSUE #90: Check and fix used bech32 prefixes
When displaying key hashes, they should be encoded in bech32 and use a correct prefix: `addr_vkh`

janmazak:
Resolves #90 .

It seems we have no more problems with bech32, hex buffer screen is used where it should, and bech32 prefixes seem correct (keys are almost always displayed as paths).


mkv-vcm (commented):
Assuming this PR should be only the "remove bech32prefix TODO" commit, it's good.

Issue #99, opened 2021-08-27
refactor certificate handling


janmazak:
``Certificate handlers and related UI machines should be refactored so that specific certificate types are handled individually. They are too disparate and the current code is a mess.

It would be best if most of the code was moved outside of `signTx.c` into a new file `signTxCertificate.c`.

See also
```
	// TODO refactor --- does it make sense to process different certificate types entirely separately?
	// or perhaps group registration with deregistration?
	// notice that _parseCertificateData and _addCertificateDataToTx already do a big switch on cert type
```

PR #103, opened 2021-08-30
refactor certificates

janmazak:


janmazak:
Resolves #99

janmazak:
Unfortunately, it increases code size by about 900 bytes. Merging registration and deregistration might mitigate it.

mkv-vcm (approved):

PR #104, opened 2021-08-30
Refactor security policies

janmazak:


src/keyDerivation.c
```diff
@@ -11,42 +11,20 @@
...
 )
 {
-	if (!_isPathAllowed(pathSpec)) {
-		THROW(ERR_INVALID_BIP44_PATH);
-	}
 	// Sanity check
 	ASSERT(pathSpec->length <= ARRAY_LEN(pathSpec->path));
 
+	// if the path is invalid, it's a bug in previous validation
+	ASSERT(policyForDerivePrivateKey(pathSpec) != POLICY_DENY);
```
mkv-vcm, line 26: ensure_not_denied? Just for consistency? Or we don't want to throw, we want to assert here?

mkv-vcm (commented):
How do we decide what is a security policy and what is a data validation?

src/keyDerivation.c
janmazak: Yes, just an assert, the path should have been validated previously.

janmazak:
Generally, the difference between validation and security policies is that data deemed valid could be serialized into a valid tx, but it is not safe to do so from the perspective of the HW wallet user (e.g. he is not aware what he is signing). 

This is not that straightforward to implement in Ledger, so in some places we might deviate from this basic principle.

mkv-vcm (approved):

PR #105, opened 2021-08-31
enforce canonical cbor map key ordering

janmazak:


src/signTx.h
```diff
@@ -85,6 +85,7 @@ typedef struct {
 typedef struct {
 	stake_credential_t stakeCredential;
 	uint64_t amount;
+	uint8_t previousRewardAccount[REWARD_ACCOUNT_SIZE];
```
mkv-vcm, line 88: If we only store this one in the struct and the stakeCredential is only on the stack in signTx_handleWithdrawalAPDU we can spare some static memory.

Maybe also should be renamed rewardAccountHash, and the next hash calculated onto the stack, as it's done in Mint and Output, makes for a cleaner data structure.

janmazak: The problem is UI --- we need to keep stakeCredential available to the UI machine, and the ui handler function does not (and cannot) have any arguments, so it has to be placed in global data.

mkv-vcm: Yes, I see. VCode is not that good at finding all the relevant references for a common name like stakeCredential.

mkv-vcm (approved):

PR #107, opened 2021-09-01
Fix security policies

janmazak:


mkv-vcm (approved):
Good code saving on the derive part!

src/securityPolicy.c
```diff
@@ -768,6 +721,37 @@ static inline security_policy_t multisigWitnessPolicy(const bip44_path_t* pathSp
...
+
+	default:
+		DENY();
+		break;
+	}
+}
+
+static inline security_policy_t _poolRegistrationOperatorWitnessPolicy(const bip44_path_t* path)
+{
+	switch (bip44_classifyPath(path)) {
```
gabrielKerekes, line 741: Since all the `*WitnessPolicy` functions call `bip44_classifyPath` it might make sense to call it in `policyForSignTxWitness` and pass in the path type into these functions. That way the number of calls to `bip44_classifyPath` would be reduced.

gabrielKerekes (approved):
Just one small suggestions which I'm not 100% sure about so feel free to ignore if you don't like it 😉 Otherwise (still untested) approve.

gabrielKerekes:
Additionally, please see this comment https://github.com/vacuumlabs/ledger-app-cardano-shelley/pull/98#discussion_r700768015 regarding the comment in the code I found confusing.

src/securityPolicy.c
janmazak: Since the branches where bip44_classifyPath is called are disjoint, it is called only once, and we need to path the whole path as argument anyway into the policies. Whatever other minor aspects of shuffling the code around, inlining functions, grouping function calls etc. I am leaving to the compiler optimization.

janmazak:
Merged directly to multisig-final.

Issue #66, opened 2021-04-15
Restrict access to tx parts in signTransaction instruction state


refi93:
When introducing Catalyst, we had to split the instruction state into multiple overlapping slices corresponding to separate transaction parts. This opens up new ways of "shooting ourselves in the foot" when working with those slices of state as we need to be sure to not use them simultaneously, otherwise we'd inadvertently overwrite the state for another part of the transaction.

see https://github.com/vacuumlabs/ledger-app-cardano-shelley/pull/65#discussion_r611974635 for context

janmazak:
Will be merged via #98.

PR #108, opened 2021-09-02
Instruction separation

mkv-vcm:
Proposed solution for issue #66

### DIFF
--- a/src/signTx.c
+++ b/src/signTx.c
@@ -16,25 +16,21 @@
 #include "securityPolicy.h"
 
 static ins_sign_tx_context_t* ctx = &(instructionState.signTxContext);
-static ins_sign_tx_body_context_t* txBodyCtx = &(instructionState.signTxContext.txPartCtx.body_ctx);
-static ins_sign_tx_aux_data_context_t* txAuxDataCtx = &(instructionState.signTxContext.txPartCtx.aux_data_ctx);
-static ins_sign_tx_witness_context_t* txWitnessCtx = &(instructionState.signTxContext.txPartCtx.witnesses_ctx);
 
-// TODO - maybe add an enum to the global context which would specify the active tx part?
 static inline void initTxBodyCtx()
 {
 	explicit_bzero(&ctx->txPartCtx, SIZEOF(ctx->txPartCtx));
 
 	{
 		// initialization
-		txBodyCtx->currentInput = 0;
-		txBodyCtx->currentOutput = 0;
-		txBodyCtx->currentCertificate = 0;
-		txBodyCtx->currentWithdrawal = 0;
-		txBodyCtx->feeReceived = false;
-		txBodyCtx->ttlReceived = false;
-		txBodyCtx->validityIntervalStartReceived = false;
-		txBodyCtx->mintReceived = false;
+		BODY_CTX->currentInput = 0;
+		BODY_CTX->currentOutput = 0;
+		BODY_CTX->currentCertificate = 0;
+		BODY_CTX->currentWithdrawal = 0;
+		BODY_CTX->feeReceived = false;
+		BODY_CTX->ttlReceived = false;
+		BODY_CTX->validityIntervalStartReceived = false;
+		BODY_CTX->mintReceived = false;
 	}
 }
 
@@ -42,8 +38,8 @@ static inline void initTxAuxDataCtx()
 {
 	explicit_bzero(&ctx->txPartCtx, SIZEOF(ctx->txPartCtx));
 	{
-		txAuxDataCtx->auxDataReceived = false;
-		txAuxDataCtx->auxDataType = false;
+		AUX_DATA_CTX->auxDataReceived = false;
+		AUX_DATA_CTX->auxDataType = false;
 	}
 }
 
@@ -51,7 +47,7 @@ static inline void initTxWitnessCtx()
 {
 	explicit_bzero(&ctx->txPartCtx, SIZEOF(ctx->txPartCtx));
 	{
-		txWitnessCtx->currentWitness = 0;
+		WITNESS_CTX->currentWitness = 0;
 	}
 }
 
@@ -68,7 +64,7 @@ static inline void advanceStage()
 
 		if (ctx->includeAuxData) {
 			// wait for aux data APDU(s)
-			txAuxDataCtx->auxDataReceived = false;
+			AUX_DATA_CTX->auxDataReceived = false;
 			break;
 		}
 
@@ -76,7 +72,7 @@ static inline void advanceStage()
 
 	case SIGN_STAGE_AUX_DATA:
 		if (ctx->includeAuxData) {
-			ASSERT(txAuxDataCtx->auxDataReceived);
+			ASSERT(AUX_DATA_CTX->auxDataReceived);
 		}
 
 		ctx->stage = SIGN_STAGE_BODY_INPUTS;
@@ -85,7 +81,7 @@ static inline void advanceStage()
 		{
 			// Note: make sure that everything in ctx is initialized properly
 			txHashBuilder_init(
-			        &txBodyCtx->txHashBuilder,
+			        &BODY_CTX->txHashBuilder,
 			        ctx->numInputs,
 			        ctx->numOutputs,
 			        ctx->includeTtl,
@@ -95,14 +91,14 @@ static inline void advanceStage()
 			        ctx->includeValidityIntervalStart,
 			        ctx->includeMint
 			);
-			txHashBuilder_enterInputs(&txBodyCtx->txHashBuilder);
+			txHashBuilder_enterInputs(&BODY_CTX->txHashBuilder);
 		}
 		break;
 
 	case SIGN_STAGE_BODY_INPUTS:
 		// we should have received all inputs
-		ASSERT(txBodyCtx->currentInput == ctx->numInputs);
-		txHashBuilder_enterOutputs(&txBodyCtx->txHashBuilder);
+		ASSERT(BODY_CTX->currentInput == ctx->numInputs);
+		txHashBuilder_enterOutputs(&BODY_CTX->txHashBuilder);
 		signTxOutput_init();
 		ctx->stage = SIGN_STAGE_BODY_OUTPUTS;
 
@@ -115,12 +111,12 @@ static inline void advanceStage()
 
 	case SIGN_STAGE_BODY_OUTPUTS:
 		// we should have received all outputs
-		ASSERT(txBodyCtx->currentOutput == ctx->numOutputs);
+		ASSERT(BODY_CTX->currentOutput == ctx->numOutputs);
 		ctx->stage = SIGN_STAGE_BODY_FEE;
 		break;
 
 	case SIGN_STAGE_BODY_FEE:
-		ASSERT(txBodyCtx->feeReceived);
+		ASSERT(BODY_CTX->feeReceived);
 
 		ctx->stage = SIGN_STAGE_BODY_TTL;
 
@@ -133,13 +129,13 @@ static inline void advanceStage()
 
 	case SIGN_STAGE_BODY_TTL:
 		if (ctx->includeTtl) {
-			ASSERT(txBodyCtx->ttlReceived);
+			ASSERT(BODY_CTX->ttlReceived);
 		}
 
 		ctx->stage = SIGN_STAGE_BODY_CERTIFICATES;
 
 		if (ctx->numCertificates > 0) {
-			txHashBuilder_enterCertificates(&txBodyCtx->txHashBuilder);
+			txHashBuilder_enterCertificates(&BODY_CTX->txHashBuilder);
 			break;
 		}
 
@@ -147,12 +143,12 @@ static inline void advanceStage()
 
 	case SIGN_STAGE_BODY_CERTIFICATES:
 		// we should have received all certificates
-		ASSERT(txBodyCtx->currentCertificate == ctx->numCertificates);
+		ASSERT(BODY_CTX->currentCertificate == ctx->numCertificates);
 
 		ctx->stage = SIGN_STAGE_BODY_WITHDRAWALS;
 
 		if (ctx->numWithdrawals > 0) {
-			txHashBuilder_enterWithdrawals(&txBodyCtx->txHashBuilder);
+			txHashBuilder_enterWithdrawals(&BODY_CTX->txHashBuilder);
 			break;
 		}
 
@@ -160,15 +156,13 @@ static inline void advanceStage()
 
 	case SIGN_STAGE_BODY_WITHDRAWALS:
 		// we should have received all withdrawals
-		ASSERT(txBodyCtx->currentWithdrawal == ctx->numWithdrawals);
+		ASSERT(BODY_CTX->currentWithdrawal == ctx->numWithdrawals);
 
 		if (ctx->includeAuxData) {
-			ASSERT(txAuxDataCtx->auxDataReceived);
-
 			// add auxiliary data to tx
 			TRACE("Adding auxiliary data hash to tx hash");
 			txHashBuilder_addAuxData(
-			        &txBodyCtx->txHashBuilder,
+			        &BODY_CTX->txHashBuilder,
 			        ctx->auxDataHash, SIZEOF(ctx->auxDataHash)
 			);
 		}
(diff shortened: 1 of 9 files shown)


src/signTx.c
```diff
@@ -1855,3 +1872,52 @@ void signTx_handleAPDU(
...
+ins_sign_tx_aux_data_context_t* accessAuxDataContext()
+{
+	switch (ctx->stage)
+	{
+	case SIGN_STAGE_AUX_DATA:
+	case SIGN_STAGE_AUX_DATA_CATALYST_REGISTRATION_SUBMACHINE:
+		return &(ctx->txPartCtx.aux_data_ctx);
+	
+	default:
+		return NULL;
```
janmazak, line 1885: I would say we just want to do `ASSERT(false)` here (yes, NULL can be returned to make the compiler happy) and never check the returned value in other places. The value NULL is not meaningful and there is no point in returning it, we must stop processing the tx immediately if a wrong part of the context is accessed.

src/signTx.c
```diff
@@ -1855,3 +1872,52 @@ void signTx_handleAPDU(
 	VALIDATE(subhandler != NULL, ERR_INVALID_REQUEST_PARAMETERS);
 	subhandler(p2, wireDataBuffer, wireDataSize);
 }
+
+ins_sign_tx_aux_data_context_t* accessAuxDataContext()
```
janmazak, line 1876: I'm in favor of just `auxDataCtx()`  to keep the name short. (Similarly in other places.)

src/signTx.c
```diff
@@ -133,42 +132,42 @@ static inline void advanceStage()
...
 	// intentional fallthrough
 
 	case SIGN_STAGE_BODY_WITHDRAWALS:
 		// we should have received all withdrawals
-		ASSERT(txBodyCtx->currentWithdrawal == ctx->numWithdrawals);
+		ASSERT(ACCESS_BODY_CTX()->currentWithdrawal == ctx->numWithdrawals);
 
 		if (ctx->includeAuxData) {
-			ASSERT(txAuxDataCtx->auxDataReceived);
+			// ASSERT(txAuxDataCtx->auxDataReceived);	TODO(KoMa) why is this check here? We are deep in txBody territory here!
```
janmazak, line 165: Seems like a bug, some leftover from the times where aux data were received at the end of tx body after withdrawals.

src/signTx.c
mkv-vcm: I removed the ASSERT in one of the commits because it doesn't help much when trying to debug. If you call the function directly, you get the assert in the function, and will have no idea where it comes from. Hence why I wrote the macros. If it were not C, the functions would be private and not callable at all from the outside.

src/signTx.c
mkv-vcm: Can be done, sure.

mkv-vcm: Except, it collides with the variable names, and I think the function of the function is better shown in the name as well. It makes checks. Let me think about a better naming for the macros, those are the visible part.

src/signTx.c
mkv-vcm: Removed then.

src/signTx.c
janmazak: I've dismissed that on the grounds of having enough tracing to keep one aware of what is being executed (in particular, what stage one is in, which is the determinant of whether the function fails or not). It is true that debugging would be easier if one knew precisely where the context access failed.

But dereferencing a NULL pointer is undefined behavior (not a guaranteed program crash), and I am afraid someone might use the function without the macro (impossible to guard against in C afaik, macros don't provide any scope-controlling features). This is a much bigger concern that longer debugging.

What about having two versions of `accessAuxDataContext` (one for debugging purposes, commented out, the other failing the assertion) and then AUX_DATA_CTX will only define which of them will be called?

We will also save on code size a bit (no redundant NULL checks).

Also, `AUX_DATA_CTX` does not need arguments, it can be treated as a variable when used (although it does involve a function call).

janmazak: The macro structure probably needs to be a bit more complex, but it is essentially doable (maybe you need helper macros to achieve the effect I tried to describe).

mkv-vcm: Just pushed a suggestion, what do you think about that?

src/signTx.h
```diff
@@ -166,9 +166,15 @@ ins_sign_tx_aux_data_context_t* accessAuxDataContext();
...
-#define WITNESS_CTX() (ASSERT(accessWitnessContext() != NULL), accessWitnessContext())
+#ifdef DEVEL
+#define DEBUG_ONLY_ASSERT(cond) ASSERT(cond),
+#else
+#define DEBUG_ONLY_ASSERT(cond) 
+#endif
+
+#define AUX_DATA_CTX() (DEBUG_ONLY_ASSERT(accessAuxDataContext() != NULL) accessAuxDataContext())
+#define BODY_CTX() (DEBUG_ONLY_ASSERT(accessBodyContext() != NULL) accessBodyContext())
+#define WITNESS_CTX() (DEBUG_ONLY_ASSERT(accessWitnessContext() != NULL) accessWitnessContext())
```
janmazak, line 177: I'd try to avoid "incomplete" macros (ending with an operator), e.g. like this:
```
#ifdef DEVEL
#define ASSERT_DEVEL_ONLY(cond) ASSERT(cond)
#else
#define ASSERT_DEVEL_ONLY(cond) 0
#endif

#define AUX_DATA_CTX() (ASSERT_DEVEL_ONLY(accessAuxDataContext() != NULL), accessAuxDataContext())
```
Otherwise, I like it, it seems to solve all our problems (no need to rename the access* functions).

janmazak: But maybe doing it directly is even clearer?
```
#ifdef DEVEL
#define AUX_DATA_CTX (ASSERT(accessAuxDataContext() != NULL), accessAuxDataContext())
#else
#define AUX_DATA_CTX (accessAuxDataContext())
#endif

```

mkv-vcm: The first version was the one I tried, it produces a ton of warnings about unused values (the 0), so I ditched that, but yeah, your second version is actually the good solution I think, not sure why I didn't think of that.

src/signTx.c
```diff
@@ -23,7 +23,7 @@ static inline void initTxBodyCtx()
 	explicit_bzero(&ctx->txPartCtx, SIZEOF(ctx->txPartCtx));
 
 	{
-		ins_sign_tx_body_context_t* bodyCtx = BODY_CTX();
+		ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 26: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -40,7 +40,7 @@ static inline void initTxAuxDataCtx()
 {
 	explicit_bzero(&ctx->txPartCtx, SIZEOF(ctx->txPartCtx));
 	{
-		ins_sign_tx_aux_data_context_t* auxDataCtx = AUX_DATA_CTX();
+		ins_sign_tx_aux_data_context_t* auxDataCtx = AUX_DATA_CTX;
```
janmazak, line 43: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -231,16 +227,18 @@ static inline void advanceCertificatesStateIfAppropriate()
 
 	switch (ctx->stage) {
 
-	case SIGN_STAGE_BODY_CERTIFICATES:
-		ASSERT(txBodyCtx->currentCertificate < ctx->numCertificates);
+	case SIGN_STAGE_BODY_CERTIFICATES: {
+		ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 231: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -264,12 +262,13 @@ static inline void checkForFinishedSubmachines()
 	switch (ctx->stage) {
 	case SIGN_STAGE_BODY_OUTPUTS_SUBMACHINE:
 		if (signTxOutput_isFinished()) {
+			ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 265: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -289,11 +288,12 @@ static inline void checkForFinishedSubmachines()
 		if (signTxCatalystRegistration_isFinished()) {
 			TRACE();
 			ctx->stage = SIGN_STAGE_AUX_DATA;
-			txAuxDataCtx->auxDataReceived = true;
+			ins_sign_tx_aux_data_context_t* auxDataCtx = AUX_DATA_CTX;
```
janmazak, line 291: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -581,21 +581,21 @@ static void signTx_handleAuxDataAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t
 			CHECK_STAGE(SIGN_STAGE_AUX_DATA);
 		}
 	}
-
+	ins_sign_tx_aux_data_context_t* auxDataCtx = AUX_DATA_CTX;
```
janmazak, line 584: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -663,11 +663,12 @@ static void signTx_handleInput_ui_runStep()
 	UI_STEP(HANDLE_INPUT_STEP_RESPOND) {
 		respondSuccessEmptyMsg();
 
+		ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 666: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -798,23 +799,23 @@ static void signTx_handleFeeAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wir
 		VALIDATE(p2 == P2_UNUSED, ERR_INVALID_REQUEST_PARAMETERS);
 		ASSERT(wireDataSize < BUFFER_SIZE_PARANOIA);
 	}
-
+	ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 802: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -875,24 +876,24 @@ static void signTx_handleTtlAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wir
...
```
janmazak, line 879: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -929,7 +930,8 @@ static void signTx_handleCertificate_ui_runStep()
 	UI_STEP_BEGIN(ctx->ui_step, this_fn);
 
 	UI_STEP(HANDLE_CERTIFICATE_STEP_DISPLAY_OPERATION) {
-		switch (txBodyCtx->stageData.certificate.type) {
+		ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 933: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -963,19 +965,20 @@ static void signTx_handleCertificate_ui_runStep()
 		}
 	}
 	UI_STEP(HANDLE_CERTIFICATE_STEP_DISPLAY_STAKING_KEY) {
-		switch (txBodyCtx->stageData.certificate.stakeCredential.type) {
+		ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 968: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -1031,7 +1034,8 @@ enum {
 static void signTx_handleCertificatePoolRetirement_ui_runStep()
 {
 	TRACE("UI step %d", ctx->ui_step);
-	ASSERT(txBodyCtx->stageData.certificate.type == CERTIFICATE_TYPE_STAKE_POOL_RETIREMENT);
+	ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 1037: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -1171,18 +1175,19 @@ static void _addCertificateDataToTx(
 )
 {
 	// data only added in the sub-machine, see signTxPoolRegistration.c
-	ASSERT(txBodyCtx->stageData.certificate.type != CERTIFICATE_TYPE_STAKE_POOL_REGISTRATION);
+	ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 1178: Seems we don't need this variable anymore?

src/signTx.c
```diff
@@ -1220,7 +1225,8 @@ static void signTx_handleCertificateAPDU(uint8_t p2, uint8_t* wireDataBuffer, si
 {
 	TRACE_STACK_USAGE();
 
-	ASSERT(txBodyCtx->currentCertificate < ctx->numCertificates);
+	ins_sign_tx_body_context_t* bodyCtx = BODY_CTX;
```
janmazak, line 1228: Seems we don't need this variable anymore? Probably many more times around.

src/signTx.c
```diff
@@ -1855,3 +1870,61 @@ void signTx_handleAPDU(
 	VALIDATE(subhandler != NULL, ERR_INVALID_REQUEST_PARAMETERS);
 	subhandler(p2, wireDataBuffer, wireDataSize);
 }
+
+ins_sign_tx_aux_data_context_t* accessAuxDataContext()
+{
+	switch (ctx->stage)
+	{
```
janmazak, line 1877: Formatting.

src/signTxCatalystRegistration.c
```diff
@@ -40,24 +47,22 @@ bool signTxCatalystRegistration_isFinished()
...
-		explicit_bzero(&auxDataCtx->stageContext, SIZEOF(auxDataCtx->stageContext));
-	}
-
-	auxDataHashBuilder_init(auxDataHashBuilder);
+	ins_sign_tx_aux_data_context_t* auxDataCtx = AUX_DATA_CTX;
+	explicit_bzero(&auxDataCtx->stageContext, SIZEOF(auxDataCtx->stageContext));
+	auxDataHashBuilder_init(&auxDataCtx->auxDataHashBuilder);
 
-	subctx->state = STATE_CATALYST_REGISTRATION_VOTING_KEY;
+	accessSubContext()->state = STATE_CATALYST_REGISTRATION_VOTING_KEY;
```
janmazak, line 54: Maybe a macro SUB_CTX analogous to the other ones? Or even better one with a more specific name? In other places, too.

src/signTxMint.c
```diff
@@ -7,18 +7,25 @@
 #include "textUtils.h"
 #include "securityPolicy.h"
 
-static mint_context_t* subctx = &(instructionState.signTxContext.txPartCtx.body_ctx.stageContext.mint_subctx);
+// static mint_context_t* subctx = &(instructionState.signTxContext.txPartCtx.body_ctx.stageContext.mint_subctx);
```
janmazak, line 10: These comments should probably go away.

janmazak (commented):
There's  an ongoing discussion with Ledger devs about what PR this should be included into. So DO NOT MERGE, I'll do it later.

I think we are close to done, I feel good about the current approach to the macros, and I think it is worth extending it to subcontexts.

janmazak:
Can be squashed into a single commit (that's how it will be passed on to Ledger).

src/signTx.c
mkv-vcm: It's still a function call, hidden in the macro.

src/signTxCatalystRegistration.c
mkv-vcm: I don't see the advantage. I use the macros to inject ASSERT code into the function call with usable debug info. The subcontext access happens through these macros already.

src/signTx.c
mkv-vcm: It's weird. Our formatting rules push the #if macros to the indentation level of the governed code, but remove the macro indentation that is inside an #if. Well, whatever.

src/signTx.c
janmazak: Hmm, I would not be concerned about speed here, more about readability. I am not sure what the compiler does with in the end (e.g. inlining the function call everywhere vs. storing it in a variable and calling only once) --- perhaps try it with the variable, then without, and tell me what the resulting app size is (does the -O3 gcc option affect it? --- careful, the option is already included in Makefile.defines, so it has to be removed not only in our Makefile, but also there) --- if it is the same, I'd go without the variable.

janmazak:
Will be merged via #98.

Issue #97, opened 2021-08-26
different ways of printing integers


janmazak:
We are currently using different ways of printing integers (whether signed or unsigned). It should be unified where reasonable, and we might need to add some assertions similar to
```
STATIC_ASSERT(sizeof(int) >= sizeof(blockchainIndex_t), "bad blockchainIndex_t size");
ASSERT(blockchainPointer.blockIndex <= INT32_MAX);
```
ensuring that the printf format we use, e.g. %d, can accomodate the values being printed.

janmazak:
Resolved in #98.

PR #109, opened 2021-09-06
Issue #97: additional checks for outputting 

mkv-vcm:
Just an exploratory code to see what you meant with the issue.

### DIFF
--- a/src/addressUtilsShelley.c
+++ b/src/addressUtilsShelley.c
@@ -432,32 +432,25 @@ size_t deriveAddress(const addressParams_t* addressParams, uint8_t* outBuffer, s
 	return BUFFER_SIZE_PARANOIA + 1;
 }
 
-// TODO(ppershing): this function needs to be thoroughly tested
-// on small outputSize
 void printBlockchainPointerToStr(blockchainPointer_t blockchainPointer, char* out, size_t outSize)
 {
 	ASSERT(outSize < BUFFER_SIZE_PARANOIA);
-	// We have to have space for terminating null
 
-	STATIC_ASSERT(sizeof(int) >= sizeof(blockchainIndex_t), "bad blockchainIndex_t size");
-	STATIC_ASSERT(sizeof(int) == 4, "bad int size"); // because of the checks below
-	ASSERT(blockchainPointer.blockIndex <= INT32_MAX);
-	ASSERT(blockchainPointer.txIndex <= INT32_MAX);
-	ASSERT(blockchainPointer.certificateIndex <= INT32_MAX);
+	STATIC_ASSERT(sizeof(blockchainIndex_t) <= sizeof(unsigned), "oversized type for %u");
+	STATIC_ASSERT(!IS_SIGNED(blockchainPointer.blockIndex), "signed type for %u");
+	STATIC_ASSERT(!IS_SIGNED(blockchainPointer.txIndex), "signed type for %u");
+	STATIC_ASSERT(!IS_SIGNED(blockchainPointer.certificateIndex), "signed type for %u");
 
 	ASSERT(outSize > 0);
 	snprintf(
 	        out,
 	        outSize,
-	        "(%d, %d, %d)",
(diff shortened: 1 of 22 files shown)


src/deriveNativeScriptHash.c
```diff
@@ -169,6 +173,10 @@ static void deriveScriptHash_display_ui_runStep()
 			// max possible length 85: "Requires n out of k signatures. Contains k nested scripts."
 			// where n and k is 2^32-1
 			char text[86] = {0};
+			STATIC_ASSERT(sizeof(unsigned) >= sizeof(ctx->scriptContent.requiredScripts), "oversized type for %u");
+			STATIC_ASSERT(sizeof(unsigned) >= sizeof(ctx->complexScripts[ctx->level].remainingScripts), "oversized type for %u");
+			STATIC_ASSERT(!IS_SIGNED(ctx->scriptContent.requiredScripts), "signed type for %u");
```
janmazak, line 178: I guess I'd prefer to group these by the variable that is being checked, not the check type, but don't feel strongly about it.

janmazak: Also, it might be worth to have local variables to avoid repeating `ctx->complexScripts[ctx->level].remainingScripts` and `ctx->scriptContent.requiredScripts` again and again.

src/ipUtils.c
```diff
@@ -141,7 +141,8 @@ void inet_ntop6 (const uint8_t *src, char *dst, size_t dstSize)
 			tp += strlen(tp);
 			break;
 		}
-		ASSERT((unsigned int) i < SIZEOF(words));
+		STATIC_ASSERT(sizeof(unsigned) >= sizeof(words[i]), "oversized type for %u");
```
janmazak, line 144: I think we are using `SIZEOF` instead of `sizeof` for variables (adds a pointer vs. array check, does not work for some specific array size, 4 or something).

src/addressUtilsShelley.c
```diff
@@ -439,20 +439,19 @@ void printBlockchainPointerToStr(blockchainPointer_t blockchainPointer, char* ou
 	ASSERT(outSize < BUFFER_SIZE_PARANOIA);
 	// We have to have space for terminating null
 
-	STATIC_ASSERT(sizeof(int) >= sizeof(blockchainIndex_t), "bad blockchainIndex_t size");
-	STATIC_ASSERT(sizeof(int) == 4, "bad int size"); // because of the checks below
-	ASSERT(blockchainPointer.blockIndex <= INT32_MAX);
-	ASSERT(blockchainPointer.txIndex <= INT32_MAX);
-	ASSERT(blockchainPointer.certificateIndex <= INT32_MAX);
+	STATIC_ASSERT(sizeof(unsigned) >= sizeof(blockchainIndex_t), "oversized type for %u");
```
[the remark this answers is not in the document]
mkv-vcm: Done.

src/deriveNativeScriptHash.c
mkv-vcm: Can be done, it's a taste thing, I don't feel strongly about it either.

I like to repeat the expressions, because it shows that we check the exact same thing, without obfuscating the original code. And it's compile time only.

src/ipUtils.c
mkv-vcm: I thought about that, but for printf things it doesn't really matter, you never print an array unless it's %s. Can replace it if the aesthetics would be better, no problem.

src/deriveNativeScriptHash.c
janmazak: OK, let's keep the order of checks as it is, just get rid of the long variable references.

src/ipUtils.c
mkv-vcm: Ah, SIZEOF doesn't really work for 4 byte types, since it's the same size as a pointer. Interesting, though it's smarter than that.
`(sizeof(__typeof(int[0 - (sizeof(var) == sizeof((void *)0))])) * 0)`

PR #111, opened 2021-09-07
show multisig inputs

janmazak:
@mkv-vcm: it might make some of your refactor PRs incomplete, they would need updating.


src/signTx.c
```diff
@@ -649,17 +649,51 @@ static void signTx_handleAuxDataAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t
...
 
+	sign_tx_input_data_t* input = &txBodyCtx->stageData.input;
+
 	UI_STEP_BEGIN(ctx->ui_step, this_fn);
 
+	UI_STEP(HANDLE_INPUT_STEP_SHOW_TX_HASH) {
+		char header[30];
+		explicit_bzero(header, SIZEOF(header));
+		snprintf(header, SIZEOF(header), "Input %u: tx id", txBodyCtx->currentInput + 1);
+		ASSERT(strlen(header) < SIZEOF(header));
```
mkv-vcm, line 672: We don't trust snprintf this much?

src/signTx.c
```diff
@@ -687,11 +721,7 @@ static void signTx_handleInputAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t w
 		ASSERT(wireDataSize < BUFFER_SIZE_PARANOIA);
 	}
 
-	// parsed data
-	struct {
-		uint8_t txHashBuffer[TX_HASH_LENGTH];
-		uint32_t parsedIndex;
-	} input;
+	sign_tx_input_data_t* input = &txBodyCtx->stageData.input;
```
mkv-vcm, line 724: You should have a look at #108, because these modifications will clump up into a backlog.

mkv-vcm (approved):

src/signTx.c
janmazak: We do, but this strict inequality ensures that the allocated buffer size is enough to include '\0'.

janmazak:
DO NOT MERGE

(ongoing discussion with IOHK about whether we should or shouldn't)

src/signTx.c
mkv-vcm: snprintf does that already.

> Maximum number of bytes to be used in the buffer.
> The generated string has a length of at most n-1, leaving space for the additional terminating null character.
size_t is an unsigned integral type.

janmazak: Hm, good to know. It really does, I checked nanos-secure-sdk repo. But the code of snprintf also contains
```
                        // skip '-' still buggy ...
```
so I am wary of removing the check, it does not really hurt to keep it.

janmazak: Ah, but it does not, for buffers of size 1  --- so no, we don't fully trust snprintf  :D
```
    if(str == NULL ||str_size < 2) {
      return 0;
    }

```

mkv-vcm: This snprintf implementation is so weird. And on the website they say to use C11 features and how advanced they are, while the SDK lacks these random things.

janmazak: There they go with C11 features:
https://github.com/LedgerHQ/app-cardano/pull/16
Using gnu11 produced clang analyzer warnings (tons of the same kind in nanos sdk repo mostly), so I asked them what can be done about that, and... this is the result :)

PR #112, opened 2021-09-13
introduce network id transaction body element

mkv-vcm:
Should we remove the networkId sent in the transaction init APDU?


mkv-vcm:
@janmazak so two questions:

1. element 15 is not mandatory in the tx body, but we add it all the time. Should the txHashBuilder code handle it as optional like now, or should it assume it's always there?
2. what are these "By always including the network id, we can get rid of the warning and the **related logic**" parts that we can remove? We still want _some_ checks on the network id, I don't really find much code to throw out.

src/securityPolicy.c
```diff
@@ -276,9 +276,11 @@ security_policy_t policyForSignTxInit(
 		ASSERT(false);
 	}
 
+	DENY_IF(networkId > MAXIMUM_NETWORK_ID);
```
janmazak, line 279: Why don't we use `isValidNetworkId`?

src/txHashBuilder.c
```diff
@@ -1180,10 +1183,46 @@ static void txHashBuilder_assertCanLeaveMint(tx_hash_builder_t* builder)
 	ASSERT(builder->multiassetData.remainingTokens == 0);
 }
 
+void txHashBuilder_addNetworkId(tx_hash_builder_t* builder, uint8_t networkId)
+{
+	_TRACE("state = %d", builder->state);
+
+	txHashBuilder_assertCanLeaveMint(builder);
+
+	// add fee item into the main tx body map
```
janmazak, line 1192: not fee

src/txHashBuilder.h
```diff
@@ -15,6 +15,7 @@ enum {
 	TX_BODY_KEY_AUX_DATA = 7,
 	TX_BODY_KEY_VALIDITY_INTERVAL_START = 8,
 	TX_BODY_KEY_MINT = 9,
+        TX_BODY_KEY_NETWORK_ID = 15,
```
janmazak, line 18: Indentation.

src/txHashBuilder.c
```diff
@@ -1180,10 +1183,46 @@ static void txHashBuilder_assertCanLeaveMint(tx_hash_builder_t* builder)
...
+	case TX_HASH_BUILDER_IN_MINT:
+	case TX_HASH_BUILDER_IN_VALIDITY_INTERVAL_START:
+	case TX_HASH_BUILDER_IN_AUX_DATA:
+	case TX_HASH_BUILDER_IN_WITHDRAWALS:
+	case TX_HASH_BUILDER_IN_CERTIFICATES:
+	case TX_HASH_BUILDER_IN_TTL:
+	case TX_HASH_BUILDER_IN_FEE:
+		txHashBuilder_assertCanLeaveMint(builder);
+		ASSERT(!builder->includeValidityIntervalStart);
+		break;
```
janmazak, line 1216: I'm confused by this. Is it true it will disappear if we make network id mandatory?

src/txHashBuilder.h
```diff
@@ -52,7 +53,8 @@ typedef enum {
 	TX_HASH_BUILDER_IN_MINT_TOP_LEVEL_DATA = 1010,
 	TX_HASH_BUILDER_IN_MINT_ASSET_GROUP = 1011,
 	TX_HASH_BUILDER_IN_MINT_TOKEN = 1012,
-	TX_HASH_BUILDER_FINISHED = 1100,
+	TX_HASH_BUILDER_IN_NETWORK_ID = 1100,
```
janmazak, line 56: This might need some adjustment when other tx body fields are added so that we don't have any clash. I'd leave the FINISHED value alone, though.

janmazak (commented):
1. Making it mandatory in txhashbuilder will probably allow us to save some code in the txhashbuilder state machine, won't it? (see detailed comments)
2. I agree that it won't be much code, probably only a handful of lines around the warning we are removing (I've not checked in detail yet).

So far, this only includes changes in the hash builder? And signTx changes will be done within this PR?

src/securityPolicy.c
mkv-vcm: True.

src/txHashBuilder.c
mkv-vcm: True.

src/txHashBuilder.h
mkv-vcm: Done.

src/txHashBuilder.c
mkv-vcm: Well, it would look like this: ASSERT(builder->state == TX_HASH_BUILDER_IN_NETWORK_ID); since there is no other way to leave it. Not much of a code reduction, but still some bytes.

src/txHashBuilder.h
mkv-vcm: Not sure how to make it certain it causes a merge conflict and demands attention then.

mkv-vcm:
> 2. I agree that it won't be much code, probably only a handful of lines around the warning we are removing (I've not checked in detail yet).

I think I modified all the relevant parts, but I might have missed some.

> So far, this only includes changes in the hash builder? And signTx changes will be done within this PR?

As far as I see, there is no need for much modification in signTx. That one line I added in this PR so far is all we need, really. Since it's a mandatory element and not tied to a APDU.

src/securityPolicy.c
```diff
@@ -276,7 +276,7 @@ security_policy_t policyForSignTxInit(
 		ASSERT(false);
 	}
 
-	DENY_IF(networkId > MAXIMUM_NETWORK_ID);
+	DENY_IF(!isValidNetworkId(networkId));
```
janmazak, line 279:
```suggestion
	DENY_UNLESS(isValidNetworkId(networkId));
```

src/securityPolicy.c
```diff
@@ -276,7 +276,7 @@ security_policy_t policyForSignTxInit(
...
 
 	WARN_IF(!is_tx_network_verifiable(numOutputs, numWithdrawals, txSigningMode));
```
janmazak, line 281:
```suggestion
	WARN_UNLESS(is_tx_network_verifiable(numOutputs, numWithdrawals, txSigningMode));
```

src/txHashBuilder.c
janmazak: I am mostly interested in the simplicity :)
Having `ASSERT(!builder->includeValidityIntervalStart);` in a function named `txHashBuilder_assertCanLeaveNetworkId` is just confusing to me, and it seems simpler to just get rid of it (by making network id always included) instead of thinking about why it is here, and whether it should be.

src/txHashBuilder.h
janmazak: I'm not sure either, but it seems to make sense to build all the other Alonzo features on top of this branch, and then we don't need to think about the conflict.

src/securityPolicy.c
mkv-vcm: Done.

src/securityPolicy.c
mkv-vcm: Done.

src/txHashBuilder.c
mkv-vcm: There are already examples for this in other mandatory elements, like Fee. I think some kind of assert is needed, so the Confirm state can check if anything is missing.

src/txHashBuilder.h
mkv-vcm: Yeah, just occurred to me while finishing another alonso feature. Will rebase and rework that after the standup.

src/securityPolicy.c
```diff
@@ -276,9 +276,11 @@ security_policy_t policyForSignTxInit(
 		ASSERT(false);
 	}
 
-	WARN_IF(!is_tx_network_verifiable(numOutputs, numWithdrawals, txSigningMode));
+	DENY_UNLESS(isValidNetworkId(networkId));
 
-	WARN_IF(networkId != MAINNET_NETWORK_ID);
+	WARN_UNLESS(is_tx_network_verifiable(numOutputs, numWithdrawals, txSigningMode));
```
janmazak, line 281: We aimed to get rid of `is_tx_network_verifiable` --- if tx body contains network id explicitly, we don't care whether it is included in ouputs or wihdrawals.

Also, the warning
```
			// technically, no withdrawals/pool reg. certificate as well, but the UI message would be too long
			ui_displayPaginatedText(
			        header,
			        "no outputs, cannot verify network id",
			        this_fn
			);

```
in signTx.c is no longer needed --- that's the "related logic" I mentioned before.

janmazak (changes requested):

src/securityPolicy.c
mkv-vcm: Ah yes, since we always know what the networkId is, or at least code it into the tx. Ok, done!

src/securityPolicy.c
```diff
@@ -239,7 +218,6 @@ security_policy_t policyForSignTxInit(
         uint8_t networkId,
         uint32_t protocolMagic,
         uint16_t numInputs MARK_UNUSED,
-        uint16_t numOutputs,
```
janmazak, line 242: I'd say we should either include all the numbers here (and mark them MARK_UNUSED if appropriate), or use just those that are relevant (I have a slight preference for this), i.e. we can remove numInputs from params.

src/signTx.c
```diff
@@ -338,20 +338,11 @@ static void signTx_handleInit_ui_runStep()
 		        "New script transaction" :
```
janmazak, line 338: Not for this PR, but when renaming the singing mode SCRIPT -> MULTISIG, we want to change this string --- please note it somewhere so that we do not forget. If "New multisig transaction" is too long to fit on the display, I'd go with "Multisig transaction".

src/securityPolicy.c
mkv-vcm: I absolutely on the side of removing unnecessary parameters.

src/signTx.c
mkv-vcm: Will try.

src/txHashBuilder.c
```diff
@@ -1180,10 +1183,46 @@ static void txHashBuilder_assertCanLeaveMint(tx_hash_builder_t* builder)
...
+
```
janmazak, line 1217:
```suggestion
```

mkv-vcm: True.

janmazak (approved):
This should be squashed into a single commit and made part of the Alonzo phase 1 branch & PR.

PR #113, opened 2021-09-16
include script data hash in transactions

mkv-vcm:
Is this how it should work?


src/signTx.c
```diff
@@ -1581,6 +1593,77 @@ static void signTx_handleMintAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
...
+		BODY_CTX->scriptDataHashReceived = true;
+	}
+
+	{
+		// add to tx
+		TRACE("Adding script data hash to tx hash");
+		txHashBuilder_addScriptDataHash(&BODY_CTX->txHashBuilder, BODY_CTX->stageData.scriptDataHash, SIZEOF(BODY_CTX->stageData.scriptDataHash));
+	}
+
+	security_policy_t policy = POLICY_SHOW_BEFORE_RESPONSE;//policyForSignTxScriptDataHash(ctx->commonTxData.txSigningMode, BODY_CTX->stageData.fee);
```
mkv-vcm, line 1649: Should we have a simple policy function for this?

janmazak: Yes. We want to keep all the decisions on security policies in a single place (as much as possible).  Also, the policy should receive singing mode as an argument, and deny script data hash for pool registrations.

src/signTx.c
```diff
@@ -193,12 +194,21 @@ static inline void advanceStage()
 		if (ctx->includeMint) {
 			ASSERT(BODY_CTX->mintReceived);
 		}
+		ctx->stage = SIGN_STAGE_BODY_SCRIPT_DATA_HASH;
+		if (ctx->includeScriptDataHash) {
+			break;
+		}
+
```
janmazak, line 201:
```suggestion

	// intentional fallthrough

```

src/signTx.c
```diff
@@ -1576,6 +1590,77 @@ static void signTx_handleMintAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
...
+		VALIDATE(view_remainingSize(&view) == 0, ERR_INVALID_DATA);
+
+		BODY_CTX->scriptDataHashReceived = true;
+	}
+
+	{
+		// add to tx
+		TRACE("Adding script data hash to tx hash");
+		txHashBuilder_addScriptDataHash(&BODY_CTX->txHashBuilder, BODY_CTX->stageData.scriptDataHash, SIZEOF(BODY_CTX->stageData.scriptDataHash));
+	}
```
janmazak, line 1644: Perhaps I'd move this after we have checked the security policy. However, it is apparently not done consistently in APDU handlers, so let's create a github issue and decide later. (And it is not really important for security because tx hash builder computations do not involve key derivation.)

src/signTx.c
```diff
@@ -1837,7 +1923,8 @@ void signTx_handleAPDU(
 	case SIGN_STAGE_BODY_WITHDRAWALS:
 	case SIGN_STAGE_BODY_VALIDITY_INTERVAL:
 	case SIGN_STAGE_BODY_MINT:
-	case SIGN_STAGE_BODY_MINT_SUBMACHINE: {
+	case SIGN_STAGE_BODY_MINT_SUBMACHINE:
+	case SIGN_STAGE_BODY_SCRIPT_DATA_HASH: {
 		explicit_bzero(&BODY_CTX->stageData, SIZEOF(BODY_CTX->stageData));
 		break;
 	}
```
janmazak, line 1930: Above this, there is a TODO I don't understand --- it seems to me that we still need to zero stageData in between the stages.

src/txHashBuilder.c
```diff
@@ -1183,25 +1187,31 @@ static void txHashBuilder_assertCanLeaveMint(tx_hash_builder_t* builder)
 	ASSERT(builder->multiassetData.remainingTokens == 0);
```
janmazak, line 1187: I'd say that these two asserts should only be done for `case TX_HASH_BUILDER_IN_MINT:`?
If I am right, the change should be made in multisig already.

janmazak (commented):
At some point, we need to update the documentation on APDU format.

src/txHashBuilder.c
mkv-vcm: I assume the data is zeroed out anyways, so if no multiassets were used (in mint or output), they should remain zero, unless pool certificate, in which case those two should be zero. But yes, logically this check is only needed in the mint case.

src/signTx.c
mkv-vcm: Well, it's not my TODO so I'm not sure what the writer thought, but I agree, the way issue #66 was solved we still need to zero out the stageData.

src/txHashBuilder.c
janmazak: It might be zeroed, but it might be not --- it's wrong to make assumptions about data that are in some union possibly used for a different purpose at the moment.

src/signTx.c
mkv-vcm: True.

src/signTx.c
mkv-vcm: #118

src/signTx.c
mkv-vcm: Removed the TODO.

src/txHashBuilder.c
mkv-vcm: I agree, done.

src/txHashBuilder.c
```diff
@@ -1163,6 +1167,8 @@ static void txHashBuilder_assertCanLeaveMint(tx_hash_builder_t* builder)
 
 	switch (builder->state) {
 	case TX_HASH_BUILDER_IN_MINT:
+		ASSERT(builder->multiassetData.remainingAssetGroups == 0);
+		ASSERT(builder->multiassetData.remainingTokens == 0);
```
janmazak, line 1171: Hasn't this been done in another PR?... I don't see a reason to make this change when adding script data hash to tx body.

janmazak (approved):
Everything apart from the issue mentioned in the comment (changes in txHashBuilder_assertCanLeaveMint, I expect they are made in another PR or can be included in a separate commit) should be squashed into a single commit and included in Alonzo phase 1.

src/txHashBuilder.c
janmazak: I am adding this small change (moving asserts into this place) to `multisig-final-update-to-be-tested-yet`, we want to use them in multisig-final.

mkv-vcm: Well, this was moved with the alonso changes that introduced multiple steps after mint. But yeah, can be moved.

PR #114, opened 2021-09-17
allow transactions with 0 witnesses to run

mkv-vcm:


src/signTx.c
```diff
@@ -199,7 +199,9 @@ static inline void advanceStage()
 		ctx->stage = SIGN_STAGE_WITNESSES;
 		initTxWitnessCtx();
 
-		break;
+		if (ctx->numWitnesses != 0) {
+			break;
+		}
```
janmazak, line 204: The comment "// intentional fallthrough" applies here.

src/signTx.c
```diff
@@ -199,7 +199,9 @@ static inline void advanceStage()
 		ctx->stage = SIGN_STAGE_WITNESSES;
 		initTxWitnessCtx();
 
-		break;
+		if (ctx->numWitnesses != 0) {
```
janmazak, line 202:
```suggestion
		if (ctx->numWitnesses > 0) {
```

src/signTx.c
mkv-vcm: True.

src/signTx.c
mkv-vcm: Not sure why it's better, but if you prefer that, can be done.

janmazak: Mostly just because we do it like this in other places. (probably because human are less good at processing negations?...)

janmazak:
Will be merged directly.

PR #115, opened 2021-09-22
WIP: validate owner witness

janmazak:
Just a small refactor as a start, TODOs should be resolved and the witness path should be validated against the owner given by path (must be the same).


src/securityPolicy.c
```diff
@@ -719,12 +722,15 @@ static inline security_policy_t _scriptWitnessPolicy(const bip44_path_t* path, b
 	}
 }
 
-static inline security_policy_t _poolRegistrationOwnerWitnessPolicy(const bip44_path_t* path)
+static inline security_policy_t _poolRegistrationOwnerWitnessPolicy(const bip44_path_t* path, const bip44_path_t* poolOwnerPath)
```
janmazak, line 725:
```suggestion
static inline security_policy_t _poolRegistrationOwnerWitnessPolicy(const bip44_path_t* witnessPath, const bip44_path_t* poolOwnerPath)
```

src/signTx.h
```diff
@@ -159,6 +159,9 @@ typedef struct {
 		ins_sign_tx_witness_context_t witnesses_ctx;
 	} txPartCtx;
 
+	bool poolOwnerByPath;
```
janmazak, line 162: Do we need this variable? I think tx signing mode fully determines whether `poolOwnerPath` is meaningful or not. (it makes sense to add an assert into the security policy to validate the argument it receives)

src/signTxPoolRegistration.c
```diff
@@ -806,8 +794,10 @@ static void signTxPoolRegistration_handleOwnerAPDU(uint8_t* wireDataBuffer, size
 			PRINTF("\n");
 
 			subctx->numOwnersGivenByPath++;
-			VALIDATE(subctx->numOwnersGivenByPath <= 1, ERR_INVALID_DATA);
-
+			// TODO this is more like a security policy
+			// VALIDATE(subctx->numOwnersGivenByPath <= 1, ERR_INVALID_DATA);
```
janmazak, line 798: Shouldn't these be removed?

PR #116, opened 2021-09-22
add output optional script data hash handling

mkv-vcm:


src/signTx.c
```diff
@@ -90,7 +90,7 @@ static inline void advanceStage()
 			        ctx->includeAuxData,
 			        ctx->includeValidityIntervalStart,
 			        ctx->includeMint,
-					ctx->includeScriptDataHash
+			        ctx->includeScriptDataHash
```
janmazak, line 93: The formatting error should be fixed in its respective branch (to avoid annoying merge conflicts). You are coding this alone, so you can force-push and rebase as needed.

src/txHashBuilder.c
```diff
@@ -342,18 +347,35 @@ void txHashBuilder_addOutput_tokenGroup(
 void txHashBuilder_addOutput_token(
         tx_hash_builder_t* builder,
         const uint8_t* assetNameBuffer, size_t assetNameSize,
-        uint64_t amount
+        uint64_t amount,
+        bool includeScriptDataHash
```
janmazak, line 351: Don't we store this in `tx_hash_builder_t`?

mkv-vcm: Indeed we do.

src/signTxOutput.c
```diff
@@ -532,18 +550,86 @@ static void signTxOutput_handleTokenAPDU(uint8_t* wireDataBuffer, size_t wireDat
...
+		        subctx->stateData.token.amount,
+		        subctx->includeDataHash
 		);
 		TRACE();
 	}
 
 	signTxOutput_handleToken_ui_runStep();
 }
 
+// ========================== SCRIPT DATA HASH =========================
```
janmazak, line 562: I find using "script data hash" here confusing. Officially, the thing being hashed is called "datum" (see https://iohk.io/en/research/library/papers/the-extended-utxo-model/), so I propose to call it "datum hash" everywhere, including in the UI (close, but not precisely equal, to "data_hash" used in CDDL), e.g. we use `HANDLE_DATUM_HASH_STEP_DISPLAY`. That way, no confusion with other script data hashes arise.

src/signTxOutput.c
```diff
@@ -532,18 +550,86 @@ static void signTxOutput_handleTokenAPDU(uint8_t* wireDataBuffer, size_t wireDat
...
+
+enum {
+	HANDLE_SCRIPT_DATA_HASH_STEP_DISPLAY_HASH = 3500,
```
janmazak, line 565:
```suggestion
	HANDLE_SCRIPT_DATA_HASH_STEP_DISPLAY = 3500,
```

src/txHashBuilder_test.c
```diff
@@ -160,10 +170,10 @@ static void addMultiassetOutput(tx_hash_builder_t* builder)
 	        builder,
 	        tmp, tmpSize,
 	        outputs[1].amount,
-	        2
+	        2, false
```
janmazak, line 173: I'd rather have "false" on a separate line.

src/signTxOutput.c
mkv-vcm: We can use datum, but it's called script_data_hash in the CDDL. Which is _more_ official?

janmazak: Only "data_hash" in CDDL:
```
transaction_output =
  [ address
  , amount : value
  , ? data_hash : $hash32 ; New
  ]

```
Anyway, https://docs.cardano.org/plutus/datums-redeemers also uses "datum"; CDDL is known to be slightly inconsistent with other docs from time to time, and has been modified at least once on our request because it is easy to change the CDDL (unlike the other docs).

src/signTxOutput.c
mkv-vcm: Yeah.

src/txHashBuilder_test.c
mkv-vcm: Done.

PR #117, opened 2021-09-29
rename script tx to multisig

janmazak:


janmazak:
Will be merged directly into multisig-final.

PR #119, opened 2021-10-01
multisig update

janmazak:


src/securityPolicy.c
```diff
@@ -343,17 +352,28 @@ security_policy_t policyForSignTxOutputAddressParams(
...
+	// address type and network identification
+	switch (params->type) {
+
+	case BYRON:
 		DENY_IF(params->protocolMagic != protocolMagic);
-	} else { // shelley
+		break;
+
+	case REWARD_KEY:
+	case REWARD_SCRIPT:
```
mkv-vcm, line 363: These also fall under the determineSpendingChoice, since they result in SPENDING_NONE, is it good to separate them like this? Are reward addresses this much a special case?

src/securityPolicy.c
```diff
@@ -301,16 +301,25 @@ security_policy_t policyForSignTxOutputAddressBytes(
...
+	const uint8_t addressNetworkId = getNetworkId(rawAddressBuffer[0]);
+
+	switch (addressType) {
+
+	case BYRON:
+		DENY_IF(extractProtocolMagic(rawAddressBuffer, rawAddressSize) != protocolMagic);
+		break;
+
+	case REWARD_KEY:
+	case REWARD_SCRIPT:
```
mkv-vcm, line 316: Ah, didn't know that even 3rd party addresses need this restriction.

mkv-vcm (approved):

janmazak:
Merged directly into multisig-final.

PR #120, opened 2021-10-01
alonso features before collateral

mkv-vcm:
If this is what you meant, I'll close the other alonso related PRs.


mkv-vcm:
data_hash -> datum change is pending.

PR #122, opened 2021-10-06
Alonzo

mkv-vcm:


src/signTx.c
```diff
@@ -1664,6 +1692,166 @@ static void signTx_handleScriptDataHashAPDU(uint8_t p2, uint8_t* wireDataBuffer,
...
+		VALIDATE(p2 == P2_UNUSED, ERR_INVALID_REQUEST_PARAMETERS);
+		ASSERT(wireDataSize < BUFFER_SIZE_PARANOIA);
+	}
+
+	{
+		TRACE_BUFFER(wireDataBuffer, wireDataSize);
+
+		read_view_t view = make_read_view(wireDataBuffer, wireDataBuffer + wireDataSize);
+		STATIC_ASSERT(SIZEOF(BODY_CTX->stageData.requiredSigner) == VKEY_LENGTH, "wrong vkey length");
+		view_copyWireToBuffer(BODY_CTX->stageData.requiredSigner, &view, VKEY_LENGTH);
```
janmazak, line 1824: I think it makes sense to allow derivation paths, too. HW wallet users probably think in terms of paths, not actual keys which are much harder to remember or compare.

src/securityPolicy.c
```diff
@@ -840,6 +840,46 @@ security_policy_t policyForSignTxScriptDataHash(const sign_tx_signingmode_t txSi
 	DENY(); // should not be reached
 }
 
+security_policy_t policyForSignTxCollaterals(const sign_tx_signingmode_t txSigningMode)
+{
+	switch (txSigningMode) {
+	case SIGN_TX_SIGNINGMODE_ORDINARY_TX:
+	case SIGN_TX_SIGNINGMODE_MULTISIG_TX:
+		SHOW();
```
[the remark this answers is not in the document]
mkv-vcm: I'm not aware of the wider context, but in this case, wouldn't it be easier to just push the collateral number to the input security policy function?

What do you mean by mixing stake credentials? Mixing paths and script hashes?

src/securityPolicy.c
```diff
@@ -840,6 +840,46 @@ security_policy_t policyForSignTxScriptDataHash(const sign_tx_signingmode_t txSi
...
+	}
+
+	DENY();
+}
+
+security_policy_t policyForSignTxRequiredSigners(const sign_tx_signingmode_t txSigningMode)
+{
+	switch (txSigningMode) {
+	case SIGN_TX_SIGNINGMODE_ORDINARY_TX:
+	case SIGN_TX_SIGNINGMODE_MULTISIG_TX:
```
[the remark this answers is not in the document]
mkv-vcm: Done for now.

src/signTx.c
```diff
@@ -1664,6 +1692,166 @@ static void signTx_handleScriptDataHashAPDU(uint8_t p2, uint8_t* wireDataBuffer,
...
+#	define  CASE(POLICY, UI_STEP) case POLICY: {ctx->ui_step=UI_STEP; break;}
+			CASE(POLICY_SHOW_BEFORE_RESPONSE, HANDLE_COLLATERAL_STEP_DISPLAY);
+			CASE(POLICY_ALLOW_WITHOUT_PROMPT, HANDLE_COLLATERAL_STEP_RESPOND);
+#	undef   CASE
+		default:
+			THROW(ERR_NOT_IMPLEMENTED);
+		}
+	}
+	signTx_handleCollateral_ui_runStep();
+}
```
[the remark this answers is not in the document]
mkv-vcm: Implemented a rudimentary UI change into the initTx handling.

src/signTx.c
```diff
@@ -1664,6 +1692,166 @@ static void signTx_handleScriptDataHashAPDU(uint8_t p2, uint8_t* wireDataBuffer,
...
+
+		read_view_t view = make_read_view(wireDataBuffer, wireDataBuffer + wireDataSize);
+		STATIC_ASSERT(SIZEOF(BODY_CTX->stageData.requiredSigner) == VKEY_LENGTH, "wrong vkey length");
+		view_copyWireToBuffer(BODY_CTX->stageData.requiredSigner, &view, VKEY_LENGTH);
+		VALIDATE(view_remainingSize(&view) == 0, ERR_INVALID_DATA);
+	}
+
+	{
+		// add to tx
+		TRACE("Adding collateral to tx hash");
```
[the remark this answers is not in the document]
mkv-vcm: Done.

src/signTx.c
```diff
@@ -362,6 +384,14 @@ static void signTx_handleInit_ui_runStep()
 		);
 	}
 
+	UI_STEP(HANDLE_INIT_STEP_SCRIPT_RUNNING_WARNING) {
```
janmazak, line 387: I think we want to make this more general --- there might be other types of warnings in the future. The security policy resulting in the warning depends on tx init APDU only, so we can determine the reason also here, and then choose appropriate warning msg.

I don't think we should use UI_STEP_JUMP at all here. Because if security policy determined a warning should be displayed, then we must display it, or assert failure and stop processing because we have some bug.

For now, I'd be OK with just asserting that `numCollaterals > 0` and that tx signing mode is the new one for Plutus/Alonzo; we can extend it easily in the future if there are more possible warnings.

src/signTx.c
```diff
@@ -362,6 +384,14 @@ static void signTx_handleInit_ui_runStep()
...
+		if (ctx->numCollaterals == 0) {
+			UI_STEP_JUMP(HANDLE_INIT_STEP_CONFIRM);
+		}
+		//TODO(KoMa) what should be here exactly?
+		ui_displayPaginatedText("Warning", "Script running", this_fn);
```
janmazak, line 392: I'd go with something like "WARNING: Plutus script will be run", split into two lines so that all of the message is visible. Please try if it fits on the Ledger screen and then start a slack thread in ledger-dev to coordinate the msg with Trezor.

src/securityPolicy.c
```diff
@@ -259,6 +260,9 @@ security_policy_t policyForSignTxInit(
 	WARN_IF(protocolMagic != MAINNET_PROTOCOL_MAGIC);
 
 	// Could be switched to POLICY_ALLOW_WITHOUT_PROMPT to skip initial "new transaction" question
+	if (numCollaterals != 0) {
+		WARN();
+	}
```
janmazak, line 265: Instead of this, we should explicitly deal with numCollaterals in each of the cases above (`switch (txSigningMode)`), with DENY or WARN as appropriate.

janmazak (changes requested):
I'll continue the review of this when the new tx signing mode is implemented.

src/signTx.c
mkv-vcm: Good that you brought this up here, because there WILL be more kinds of warnings, and I was already thinking how to incorporate it into the UI machine without making it overly complex or verbose. Is the UI step really the good place to make complicated warnings?
I used the jump because I assumed we can warn for a variety of reasons, since the warning step was already there and printing its text out in case of network id or protocol issues. Do we want a separate warning for those as well? This would basically copy the code of the security policy to the UI.

src/securityPolicy.c
mkv-vcm: This'll solve itself with the new signing mode I think.

src/signTx.c
mkv-vcm: Good.

src/signTx.c
janmazak: For user experience, it would be nice if we could pack the warning into a single screen. Because usually you just click through the warning when you know you want to run scripts. Then UI_STEP would be OK. 

For network id / protocol magic, I think we are displaying them in HANDLE_INIT_STEP_DISPLAY_DETAILS, do we display any warning? Maybe Trezor does. Please consult with Gabo / David, if we can make it more consistent, then we should try.

For more complex warnings, I'm not sure how to approach that, but so far I don't know of any use case for anything more than a single screen.

src/securityPolicy.c
```diff
@@ -242,11 +245,18 @@ security_policy_t policyForSignTxInit(
...
 
 	case SIGN_TX_SIGNINGMODE_ORDINARY_TX:
 	case SIGN_TX_SIGNINGMODE_MULTISIG_TX:
-		// no additional validation
+		DENY_IF(numCollaterals != 0 || numRequiredSigners != 0);
+		break;
+
+	case SIGN_TX_SIGNINGMODE_PLUTUS_TX:
+		WARN_IF(numCollaterals != 0);
+		WARN_IF(numCollaterals == 0);
```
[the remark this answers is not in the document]
davidmisiak: If I understand the UI code in `signTx.c` correctly, `WARN_IF(numCollaterals != 0)` is here only because we want to show the `Plutus script will be run` warning when both collaterals and script data hash are present in `PLUTUS_TX` signing mode. Wouldn't something like this be more semantically correct then?
```
case SIGN_TX_SIGNINGMODE_PLUTUS_TX:
	// Plutus script will not be executed without collaterals
	WARN_IF(numCollaterals == 0);

	// Plutus script will not be executed without script data hash
	WARN_UNLESS(includeScriptDataHash);

	// warn the user about Plutus script execution
	WARN();
	break;
```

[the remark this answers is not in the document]
mkv-vcm: Yeah, the collaterals > 0 warning is functionally the same as the signingmode_plutus in normal cases, so you are right.

src/signTx.c
```diff
@@ -693,45 +786,27 @@ static void signTx_handleInputAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t w
...
 
-	security_policy_t policy = policyForSignTxInput();
-	TRACE("Policy: %d", (int) policy);
-	ENSURE_NOT_DENIED(policy);
-
 	{
 		// select UI steps
 		switch (policy) {
 #	define  CASE(POLICY, UI_STEP) case POLICY: {ctx->ui_step=UI_STEP; break;}
+			CASE(POLICY_SHOW_BEFORE_RESPONSE, HANDLE_INPUT_STEP_RESPOND);
```
janmazak, line 809:
```suggestion
			CASE(POLICY_SHOW_BEFORE_RESPONSE, HANDLE_INPUT_STEP_DISPLAY);
```
Please click through the integration test without headless mode to make sure the UI looks as it is supposed to.

src/signTx.c
```diff
@@ -666,6 +735,12 @@ static void signTx_handleInput_ui_runStep()
 
 	UI_STEP_BEGIN(ctx->ui_step, this_fn);
 
+	UI_STEP(HANDLE_INPUT_STEP_DISPLAY) {
+		char headerText[18];
```
janmazak, line 739: I'd bzero this, I don't remember if the Ledger snprintf adds null terminator, and let's add an assert after snprintf about the length of headerText.

src/signTx.c
```diff
@@ -1665,6 +1779,194 @@ static void signTx_handleScriptDataHashAPDU(uint8_t p2, uint8_t* wireDataBuffer,
...
+static void signTx_handleCollateral_ui_runStep()
+{
+	TRACE("UI step %d", ctx->ui_step);
+	ui_callback_fn_t* this_fn = signTx_handleCollateral_ui_runStep;
+
+	UI_STEP_BEGIN(ctx->ui_step, this_fn);
+
+	UI_STEP(HANDLE_COLLATERAL_STEP_DISPLAY) {
+		char headerText[18];
+		snprintf(headerText, SIZEOF(headerText), "Collateral #%u", BODY_CTX->currentCollateral);
```
janmazak, line 1799: bzero and assert as for tx inputs

src/uiScreens.c
```diff
@@ -800,3 +800,28 @@ void ui_displayIpPortScreen(
 	        callback
 	);
 }
+
+void ui_displayInputScreen(
```
janmazak, line 804:
```suggestion
__noinline_due_to_stack__
void ui_displayInputScreen(
```

src/signTx.c
```diff
@@ -666,6 +735,12 @@ static void signTx_handleInput_ui_runStep()
 
 	UI_STEP_BEGIN(ctx->ui_step, this_fn);
 
+	UI_STEP(HANDLE_INPUT_STEP_DISPLAY) {
+		char headerText[18];
+		snprintf(headerText, SIZEOF(headerText), "Input #%u", BODY_CTX->currentInput);
```
janmazak, line 740: Please discuss with Rafael whether inputs are indexed from 0 or 1 in user interfaces, and check a block explorer, too. (the same for collateral)

src/signTx.c
```diff
@@ -1665,6 +1779,194 @@ static void signTx_handleScriptDataHashAPDU(uint8_t p2, uint8_t* wireDataBuffer,
...
+
+	UI_STEP_BEGIN(ctx->ui_step, this_fn);
+
+	UI_STEP(HANDLE_REQUIRED_SIGNERS_STEP_DISPLAY) {
+		switch (BODY_CTX->stageData.requiredSigner.type) {
+		case REQUIRED_SIGNER_WITH_PATH:
+			ui_displayPathScreen("Required signer", &BODY_CTX->stageData.requiredSigner.keyPath, this_fn);
+			break;
+		case REQUIRED_SIGNER_WITH_HASH:
+			ui_displayHexBufferScreen("Required signer", BODY_CTX->stageData.requiredSigner.keyHash, SIZEOF(BODY_CTX->stageData.requiredSigner.keyHash), this_fn);
```
janmazak, line 1881: I'm not sure whether to use hex, so I asked IOHK: https://input-output-rnd.slack.com/archives/G010N9UFDNH/p1637576833039200

src/signTx.h
```diff
@@ -107,11 +118,26 @@ typedef struct {
 	} stageContext;
 } ins_sign_tx_aux_data_context_t;
 
+typedef enum {
+	REQUIRED_SIGNER_WITH_PATH = 0,
+	REQUIRED_SIGNER_WITH_HASH = 1
+} sign_tx_required_signer_mode_t;
```
janmazak, line 124: I'd reuse the enum `key_reference_type_t` for this purpose.

janmazak: Btw. the comments for that type should be changed:
```
	KEY_REFERENCE_PATH = 1, // aka 'DEVICE_OWNED', the key belongs to this device
	KEY_REFERENCE_HASH = 2, // aka 'THIRD_PARTY', the key is third party
```

src/securityPolicy.c
```diff
@@ -905,6 +1047,49 @@ security_policy_t policyForSignTxScriptDataHash(const sign_tx_signingmode_t txSi
...
+	default:
+		ASSERT(false);
+	}
+
+	DENY();
+}
+
+security_policy_t policyForSignTxRequiredSigners(const sign_tx_signingmode_t txSigningMode)
+{
+	//TODO rework when more information is available
```
janmazak, line 1073: I think we can remove this.

src/signTx.c
```diff
@@ -458,20 +517,26 @@ static void signTx_handleInitAPDU(uint8_t p2, uint8_t* wireDataBuffer, size_t wi
...
 		ASSERT_TYPE(ctx->numWithdrawals, uint16_t);
 		ASSERT_TYPE(ctx->numWitnesses, uint16_t);
+		ASSERT_TYPE(ctx->numCollaterals, uint16_t);
+		ASSERT_TYPE(ctx->numRequiredSigners, uint16_t);
 		ctx->numInputs            = (uint16_t) u4be_read(wireHeader->numInputs);
 		ctx->numOutputs           = (uint16_t) u4be_read(wireHeader->numOutputs);
 		ctx->numCertificates      = (uint16_t) u4be_read(wireHeader->numCertificates);
 		ctx->numWithdrawals       = (uint16_t) u4be_read(wireHeader->numWithdrawals);
 		ctx->numWitnesses         = (uint16_t) u4be_read(wireHeader->numWitnesses);
+		ctx->numCollaterals		  = (uint16_t) u4be_read(wireHeader->numCollaterals);
```
janmazak, line 527: some formatting is off?

src/signTx.h
```diff
@@ -107,11 +118,26 @@ typedef struct {
...
+	};
+} ins_sign_tx_required_signer_t;
+
 typedef struct {
 	uint16_t currentInput;
 	uint16_t currentOutput;
 	uint16_t currentCertificate;
 	uint16_t currentWithdrawal;
+	uint16_t currentCollateral;
+	uint16_t currentRequiredSigners;
```
janmazak, line 140:
```suggestion
	uint16_t currentRequiredSigner;
```

src/txHashBuilder.c
```diff
@@ -452,6 +460,27 @@ void txHashBuilder_enterCertificates(tx_hash_builder_t* builder)
...
+	case STAKE_CREDENTIAL_KEY_HASH:
+		return KEY;
+		break;
+	case STAKE_CREDENTIAL_SCRIPT_HASH:
+		return SCRIPT;
+		break;	
+	default:
+		ASSERT(false);
+		break;
+	}
```
janmazak, line 481:
```suggestion
	switch (stakeCredentialType) {
	case STAKE_CREDENTIAL_KEY_PATH:
	case STAKE_CREDENTIAL_KEY_HASH:
		return KEY;

	case STAKE_CREDENTIAL_SCRIPT_HASH:
		return SCRIPT;
	
	default:
		ASSERT(false);
	}
```

src/txHashBuilder.c
```diff
@@ -1250,11 +1279,139 @@ static void txHashBuilder_assertCanLeaveScriptDataHash(tx_hash_builder_t* builde
...
 }
 
-void txHashBuilder_addNetworkId(tx_hash_builder_t* builder, uint8_t networkId)
+void txHashBuilder_enterCollaterals(tx_hash_builder_t* builder)
 {
 	_TRACE("state = %d", builder->state);
 
 	txHashBuilder_assertCanLeaveScriptDataHash(builder);
+	{
+		// Enter inputs
```
janmazak, line 1288:
```suggestion
		// Enter collateral inputs
```

src/txHashBuilder.c
```diff
@@ -1250,11 +1279,139 @@ static void txHashBuilder_assertCanLeaveScriptDataHash(tx_hash_builder_t* builde
...
+	ASSERT(builder->remainingCollaterals > 0);
+	builder->remainingCollaterals--;
+	// Array(2)[
+	//    Bytes[hash],
+	//    Unsigned[index]
+	// ]
+	{
+		BUILDER_APPEND_CBOR(CBOR_TYPE_ARRAY, 2);
+		{
+			ASSERT(utxoHashSize == 32);
```
janmazak, line 1314:
```suggestion
			ASSERT(utxoHashSize == TX_HASH_LENGTH);
```

src/txHashBuilder.c
```diff
@@ -1250,11 +1279,139 @@ static void txHashBuilder_assertCanLeaveScriptDataHash(tx_hash_builder_t* builde
...
+static void txHashBuilder_assertCanLeaveCollaterals(tx_hash_builder_t* builder)
+{
+	_TRACE("state = %u", builder->state);
+
+	switch (builder->state) {
+	case TX_HASH_BUILDER_IN_COLLATERALS:
+		ASSERT(builder->remainingCollaterals == 0);
+		break;
+
+	case TX_HASH_BUILDER_IN_SCRIPT_HASH_DATA:
```
janmazak, line 1333: Perhaps 
```suggestion
	case TX_HASH_BUILDER_IN_SCRIPT_DATA_HASH:
```

src/txHashBuilder.c
```diff
@@ -1250,11 +1279,139 @@ static void txHashBuilder_assertCanLeaveScriptDataHash(tx_hash_builder_t* builde
...
+	}
+}
+
+void txHashBuilder_enterRequiredSigners(tx_hash_builder_t* builder)
+{
+	_TRACE("state = %d", builder->state);
+
+	txHashBuilder_assertCanLeaveCollaterals(builder);
+	{
+		// Enter inputs
```
janmazak, line 1355:
```suggestion
		// Enter required signers
```

src/txHashBuilder_test.c
```diff
@@ -371,7 +371,9 @@ void run_txHashBuilder_test()
 	                   true, // metadata
 	                   true, // validity interval start
 	                   true, // mint
-	                   false // script hash data
+	                   false, // script hash data
+	                   0,	//TODO(KoMa) testing
+	                   0	//TODO(KoMa) testing
```
janmazak, line 376: What about this?

janmazak (changes requested):

src/securityPolicy.c
```diff
@@ -499,20 +530,62 @@ security_policy_t policyForSignTxCertificateStaking(
 		ASSERT(false);
 	}
 
-	switch (txSigningMode) {
-	case SIGN_TX_SIGNINGMODE_ORDINARY_TX:
-		DENY_UNLESS(stakeCredential->type == STAKE_CREDENTIAL_KEY_PATH);
+	switch (stakeCredential->type) {
```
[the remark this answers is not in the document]
janmazak: Similarly for withdrawals.

src/signTx.c
```diff
@@ -975,6 +1050,14 @@ static void signTx_handleCertificate_ui_runStep()
 			        this_fn
 			);
 			break;
+		case STAKE_CREDENTIAL_KEY_HASH:
+			ui_displayHexBufferScreen(
```
janmazak, line 1054: We should use bech32 with "stake_vkh" prefix.

janmazak: This is being changed in Trezor, too. Perhaps make a list of UI changes and consult with David that we display everything consistently with Trezor.

src/securityPolicy.c
```diff
@@ -499,20 +530,62 @@ security_policy_t policyForSignTxCertificateStaking(
...
+			DENY();
+			break;
+
+		default:
+			ASSERT(false);
+			break;
+		}
 		break;
+
+	case STAKE_CREDENTIAL_KEY_HASH:
```
janmazak, line 568: I'd put KEY_HASH after KEY_PATH and before SCRIPT_HASH everywhere (security policy, ui screen or wherever it is used).

src/signTx.c
```diff
@@ -349,26 +382,49 @@ static void signTx_handleInit_ui_runStep()
...
 	UI_STEP_BEGIN(ctx->ui_step, this_fn);
 
-	UI_STEP(HANDLE_INIT_STEP_DISPLAY_DETAILS) {
-		char* header =
-		        (ctx->commonTxData.txSigningMode == SIGN_TX_SIGNINGMODE_MULTISIG_TX) ?
-		        "Multisig transaction" :
-		        "New transaction";
+	UI_STEP(HANDLE_INIT_STEP_PROMPT_TYPE) {
+		char bodyTxt[SIZEOF("Pool operator") + 1 + SIZEOF("transaction?") + 1];
+		snprintf(bodyTxt, SIZEOF(bodyTxt), "%s transaction?", ((const char*)PIC(uiSigningModeName[ctx->commonTxData.txSigningMode - SIGN_TX_SIGNINGMODE_ORDINARY_TX])));
```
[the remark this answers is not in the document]
mkv-vcm: Added an ASSERT for the index, good idea actually.

Pursuing an avenue for UI testing might be warranted, I'll open an issue about it.

src/txHashBuilder_test.c
mkv-vcm: I already disabled the self tests, for the lack of space on the device, I'm not sure there's much point to improve on them at this point.

src/securityPolicy.c
mkv-vcm: I'd also do that, if it's not the case somewhere, that's an error.

src/signTx.c
mkv-vcm: I'm collecting all of them and will do them all after comparing.

src/signTx.c
```diff
@@ -349,26 +382,56 @@ static void signTx_handleInit_ui_runStep()
...
 
-	UI_STEP(HANDLE_INIT_STEP_DISPLAY_DETAILS) {
-		char* header =
-		        (ctx->commonTxData.txSigningMode == SIGN_TX_SIGNINGMODE_MULTISIG_TX) ?
-		        "Multisig transaction" :
-		        "New transaction";
+	UI_STEP(HANDLE_INIT_STEP_PROMPT_SIGNINGMODE) {
+		char bodyTxt[SIZEOF("Pool operator") + 1 + SIZEOF("transaction?") + 1];
+		const uint32_t signingModeStrIndex = ctx->commonTxData.txSigningMode - SIGN_TX_SIGNINGMODE_ORDINARY_TX;
+		ASSERT(signingModeStrIndex < ARRAY_LEN(uiSigningModeName));
```
janmazak, line 388:
```suggestion
		ASSERT(signingModeStrIndex >= 0);
		ASSERT(signingModeStrIndex < ARRAY_LEN(uiSigningModeName));
```

src/signTx.c
```diff
@@ -349,26 +382,56 @@ static void signTx_handleInit_ui_runStep()
...
+		ui_displayPrompt(
+		        "Start new",
+		        bodyTxt,
+		        this_fn,
+		        respond_with_user_reject
+		);
+	}
+
+	UI_STEP(HANDLE_INIT_STEP_DISPLAY_NETWORK_DETAILS) {
+		const bool usualNetworkId = ctx->commonTxData.networkId == MAINNET_NETWORK_ID || ctx->commonTxData.networkId == TESTNET_NETWORK_ID;
```
janmazak, line 399:
```suggestion
		const bool usualNetworkId =
			ctx->commonTxData.networkId == MAINNET_NETWORK_ID ||
			ctx->commonTxData.networkId == TESTNET_NETWORK_ID;
```
The long line does not fit on my screen.

src/signTx.c
```diff
@@ -349,26 +382,56 @@ static void signTx_handleInit_ui_runStep()
...
+	UI_STEP(HANDLE_INIT_STEP_SCRIPT_RUNNING_WARNING) {
+		if (ctx->numCollaterals == 0) {
+			UI_STEP_JUMP(HANDLE_INIT_STEP_NO_COLLATERALS_WARNING);
+		}
+		ui_displayPaginatedText("WARNING:", "Plutus script will be run", this_fn);
+	}
+
+	UI_STEP(HANDLE_INIT_STEP_NO_COLLATERALS_WARNING) {
+		const bool noCollateralsIsUnusual = ctx->commonTxData.txSigningMode == SIGN_TX_SIGNINGMODE_PLUTUS_TX;
+		if (!noCollateralsIsUnusual || ctx->numCollaterals > 0) {
```
janmazak, line 423:
```suggestion
		if (!collateralExpected || ctx->numCollaterals > 0) {
```
`!noCollateralsIsUnusual` sound like a triple-negation, so is incomprehensible.

src/signTx.c
```diff
@@ -349,26 +382,56 @@ static void signTx_handleInit_ui_runStep()
...
+			UI_STEP_JUMP(HANDLE_INIT_STEP_NO_SCRIPT_DATA_HASH_WARNING);
+		}
+		ui_displayPaginatedText("WARNING:", "No collaterals given for Plutus transaction", this_fn);
+	}
+
+	UI_STEP(HANDLE_INIT_STEP_NO_SCRIPT_DATA_HASH_WARNING) {
+		const bool noScriptDataHashIsUnusual = ctx->commonTxData.txSigningMode == SIGN_TX_SIGNINGMODE_PLUTUS_TX;
```
janmazak, line 430:
```suggestion
		const bool scriptDataHashExpected = ctx->commonTxData.txSigningMode == SIGN_TX_SIGNINGMODE_PLUTUS_TX;
```

src/txHashBuilder.c
```diff
@@ -452,6 +460,26 @@ void txHashBuilder_enterCertificates(tx_hash_builder_t* builder)
...
+{
+	enum {
+		KEY = 0,
+		SCRIPT = 1
+	};
+	switch (stakeCredentialType) {
+	case STAKE_CREDENTIAL_KEY_PATH:
+	case STAKE_CREDENTIAL_KEY_HASH:
+		return KEY;
+		break;
```
janmazak, line 473: `break`s are unnecessary here. I'd consider returning -1 in `default` after the assert (don't we get a warning for a missing return statement?).

src/txHashBuilder_test.c
janmazak: I'm OK with not improving them further, but please make the comment more detailed (e.g. "collaterals not used in this basic test" or something). That would allow to remove the TODO.

janmazak (commented):
There are some comments from previous review that github was hiding, perhaps you have not seen them yet?

src/signTx.c
```diff
@@ -349,26 +382,56 @@ static void signTx_handleInit_ui_runStep()
 
 	UI_STEP_BEGIN(ctx->ui_step, this_fn);
 
-	UI_STEP(HANDLE_INIT_STEP_DISPLAY_DETAILS) {
-		char* header =
-		        (ctx->commonTxData.txSigningMode == SIGN_TX_SIGNINGMODE_MULTISIG_TX) ?
-		        "Multisig transaction" :
-		        "New transaction";
+	UI_STEP(HANDLE_INIT_STEP_PROMPT_SIGNINGMODE) {
+		char bodyTxt[SIZEOF("Pool operator") + 1 + SIZEOF("transaction?") + 1];
```
janmazak, line 386: I'd consider a bit different approach:
1. allocate a fixed-length buffer (large enough, perhaps 30, even better if you used something corresponding to the screen max line length, but the font is probably not fixed-width)
2. write into the buffer
3. assert that the text written into the buffer does not fill the entire buffer (there is space left even after '\0', i.e. the allocated size is sufficient)
What do you think?

janmazak (commented):
Take a look at
https://github.com/vacuumlabs/ledger-app-cardano-shelley/pull/133
if you agree, merge it here.

src/signTx.c
mkv-vcm: It's an unsigned integer, why do we want to assert it being non-negative?

src/txHashBuilder.c
mkv-vcm: We use a _lot_ of unnecessary breaks in switches. Is there some rule for this we should adhere to?

src/signTx.c
mkv-vcm: But that is what is happening here. bodxTxt is a fixed sized buffer, and snprintf makes sure that we don't overwrite it, including the 0 at the end.

src/txHashBuilder.c
janmazak: Not really a rule. If there is an obvious return (i.e. not hidden in a macro like in security policies), there is no need for `break`; otherwise I feel safer if break is used.

src/signTx.c
janmazak: I overlooked that somehow, no need for that assert.

src/signTx.c
janmazak: My idea was that you not only want to assert that buffer was not oveflowed, but also that all the text that is supposed to be printed is actually printed to the buffer (and that can be assured by requiring some empty space left even after \0) --- we don't want a situation in the future that a tx description is cut off, it could lead to confusion / misunderstanding / loss of funds (that's probably too much of a stretch, but we are purposefully paranoid :) ).

mkv-vcm: I see, that makes sense. Missed that detail in your first comment.

src/securityPolicy.c
```diff
@@ -783,6 +946,28 @@ static inline security_policy_t _scriptWitnessPolicy(const bip44_path_t* path, b
...
+	case PATH_POOL_COLD_KEY:
+	case PATH_MULTISIG_SPENDING_KEY:
+	case PATH_MULTISIG_STAKING_KEY:
+		WARN_UNLESS(bip44_isPathReasonable(path));
+		SHOW();
+		break;
+
+	case PATH_MINT_KEY:
+		DENY_UNLESS(mintPresent);
+		SHOW();
```
davidmisiak, line 963: I believe `PATH_POOL_COLD_KEY` should be forbidden.

Also, on Trezor we are allowing `PATH_MINT_KEY` in `PLUTUS_SIGNING_MODE` even without `mintPresent`. Not sure whether we agreed on this or the opposite, but we should make it consistent.

And maybe a `break;` missing here (also in `_multisigWitnessPolicy`)?

src/securityPolicy.c
davidmisiak: Maybe the required signers paths should be validated? (I know it's not necessary, since witness paths are validated, but maybe this is what Marton meant by the comment.)

davidmisiak (commented):
I checked the changes, focusing on consistency with Trezor and the spec doc.

src/securityPolicy.c
```diff
@@ -233,25 +218,78 @@ security_policy_t policyForShowDeriveAddress(const addressParams_t* addressParam
 	return _policyForDeriveAddress(addressParams, POLICY_SHOW_BEFORE_RESPONSE);
 }
 
+bool isNetworkUsual(uint32_t networkId, uint32_t protocolMagic)
+{
+	const bool usualNetworkId =
+	        networkId == MAINNET_NETWORK_ID ||
+	        networkId == TESTNET_NETWORK_ID;
+	const bool usualProtocolMagic = (protocolMagic == MAINNET_PROTOCOL_MAGIC);
+	return usualNetworkId && usualProtocolMagic;
```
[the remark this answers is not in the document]
janmazak: @gabrielKerekes Do you remember details on what protocol magic is used on testnet and mainnet? Let's write it down as a code comment.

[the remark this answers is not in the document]
gabrielKerekes: I think we should also consider testnet protocol magic as usual if it is usual i.e. if it belongs to the "official" testnet - we have to make sure what that value is though, see discussion in Slack.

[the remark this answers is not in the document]
janmazak: should be dealt with in e3bd1586e29e49c5db421787252f8dd699bdbfc3

src/securityPolicy.c
janmazak: 5500f432764c70b1ca703abd95ae6b85af78bac0

src/securityPolicy.c
janmazak: PATH_POOL_COLD_KEY: yes.

PATH_MINT_KEY: I'd say it some kind of misuse to sign with a mint key without a mint present, but it's hard to foresee what Plutus script authors are going to come up with. I'll allow it on Ledger to be consistent with Trezor.

src/securityPolicy.c
```diff
@@ -266,47 +304,79 @@ security_policy_t policyForSignTxInit(
...
+	DENY();
 }
 
 // For each transaction (third-party) address output
 security_policy_t policyForSignTxOutputAddressBytes(
         sign_tx_signingmode_t txSigningMode,
         const uint8_t* rawAddressBuffer, size_t rawAddressSize,
-        const uint8_t networkId, const uint32_t protocolMagic
+        const uint8_t networkId, const uint32_t protocolMagic,
+        bool includeDatumHash
```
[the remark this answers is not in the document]
janmazak: 08f45cd65640185805af394c90b7691fdbcb1f10

src/securityPolicy.c
```diff
@@ -370,23 +449,50 @@ security_policy_t policyForSignTxOutputAddressParams(
...
 		DENY();
 		break;
 	}
 
+	case SIGN_TX_SIGNINGMODE_PLUTUS_TX: {
+		// the output could affect script validation so it must not be entirely hidden
+		// Note: if we relax this, some of the above restrictions may apply
+		SHOW();
+		break;
+	}
```
davidmisiak, line 494: Just to make sure this is a conscious decision - do we really want to call the device-owned outputs in Plutus transactions "change" (since it's not technically a change)?

davidmisiak (approved):
I checked mainly the Alonzo-related logic (e.g. not the changes in `WRITE` macro and such, I don't have enough context for that). I haven't done any testing.

src/securityPolicy.c
janmazak: I think "change output" has some tradition and this more or less corresponds to the definition 
https://cips.cardano.org/cips/cip2/#changeoutput
even for Plutus. [It is hard to precisely define what change is, e.g. when part of the output comes straight from the input exclusively cotrolled by the HW wallet and another part from some script-controlled stuff. Also, one might want to write a script that controls his spending from his HW wallet, and then fund the script address from the wallet, in which case we are totally talking about "change" when spending the script-controlled money, at least in the economic sense.]

I am quite sure "device-owned" is more confusing for the users. I'd be willing to consider other options, but I am not aware of anything better than "change output".

src/signTx.c
```diff
@@ -1604,6 +2041,14 @@ static void signTx_handleConfirm_ui_runStep()
 
 	UI_STEP_BEGIN(ctx->ui_step, this_fn);
 
+	UI_STEP(HANDLE_CONFIRM_STEP_TXID) {
+		ASSERT(ctx->commonTxData.txSigningMode == SIGN_TX_SIGNINGMODE_PLUTUS_TX);
+		ui_displayHexBufferScreen(
+		        "Tx body hash",
```
davidmisiak, line 2047: Why did you change this from `Transaction id` to `Tx body hash`? I thought transaction id is the term used by public.

src/signTxOutput.c
```diff
@@ -120,6 +135,13 @@ static void signTx_handleOutput_address_ui_runStep()
 
 	UI_STEP_BEGIN(subctx->ui_step, this_fn);
 
+	UI_STEP(HANDLE_OUTPUT_ADDRESS_BYTES_STEP_WARNING_DATUM) {
+		ui_displayPaginatedText(
+		        "WARNING: output",
+		        "missing datum (could be required)",
```
davidmisiak, line 141: I'm not sure whether this wording is strong enough -- "could be required" sounds like "the tx could be refused by the blockchain if not provided" to me, whereas the reality is that the tx will be accepted and the UTXO will be unspendable (!).

src/signTx.c
janmazak: At one point, I was not sure whether tx id really is equal to tx body hash, and made this change to remind me of it. Apparently it got committed into an inappropriate branch.

src/signTxOutput.c
janmazak: What about
```
		        "WARNING: output",
		        "could be unspendable due to missing datum",
```

davidmisiak: That sounds good

PR #123, opened 2021-10-13
implement single account mode

mkv-vcm:
Should checks be included with the security policies?
Are the Byron-Shelley check correct?
Should we spare a bunch of bytes by compressing down the stored data to just three pieces (<is it byron>, <account number>, <is it stored>)?

### DIFF
--- a/src/securityPolicy.c
+++ b/src/securityPolicy.c
@@ -2,6 +2,7 @@
 #include "addressUtilsByron.h"
 #include "bip44.h"
 #include "cardano.h"
+#include "signTxUtils.h"
 
 #include "securityPolicy.h"
 
@@ -374,6 +375,7 @@ security_policy_t policyForSignTxOutputAddressParams(
 	case SIGN_TX_SIGNINGMODE_POOL_REGISTRATION_OPERATOR:
 	case SIGN_TX_SIGNINGMODE_ORDINARY_TX: {
 		DENY_UNLESS(determineSpendingChoice(params->type) == SPENDING_PATH);
+		DENY_IF(violatesSingleAccountOrStoreIt(&params->spendingKeyPath));
 		SHOW_UNLESS(is_standard_base_address(params));
 		ALLOW();
 		break;
@@ -513,6 +515,7 @@ security_policy_t policyForSignTxCertificateStaking(
 	case SIGN_TX_SIGNINGMODE_ORDINARY_TX:
 		DENY_UNLESS(stakeCredential->type == STAKE_CREDENTIAL_KEY_PATH);
 		DENY_UNLESS(bip44_isOrdinaryStakingKeyPath(&stakeCredential->keyPath));
+		DENY_IF(violatesSingleAccountOrStoreIt(&stakeCredential->keyPath));
 		break;
 	case SIGN_TX_SIGNINGMODE_MULTISIG_TX:
 		DENY_UNLESS(stakeCredential->type == STAKE_CREDENTIAL_SCRIPT_HASH);
@@ -615,6 +618,7 @@ security_policy_t policyForSignTxStakePoolRegistrationOwner(
 {
 	if (owner->keyReferenceType == KEY_REFERENCE_PATH) {
 		DENY_UNLESS(bip44_isOrdinaryStakingKeyPath(&owner->path));
+		DENY_IF(violatesSingleAccountOrStoreIt(&owner->path));
 	}
 
 	switch (txSigningMode) {
@@ -684,6 +688,7 @@ security_policy_t policyForSignTxWithdrawal(
 	case SIGN_TX_SIGNINGMODE_ORDINARY_TX:
 		DENY_UNLESS(stakeCredential->type == STAKE_CREDENTIAL_KEY_PATH);
 		DENY_UNLESS(bip44_isOrdinaryStakingKeyPath(&stakeCredential->keyPath));
+		DENY_IF(violatesSingleAccountOrStoreIt(&stakeCredential->keyPath));
 		SHOW();
 		break;
 
@@ -704,9 +709,14 @@ static inline security_policy_t _ordinaryWitnessPolicy(const bip44_path_t* path,
 	switch (bip44_classifyPath(path)) {
 	case PATH_ORDINARY_SPENDING_KEY:
 	case PATH_ORDINARY_STAKING_KEY:
-	case PATH_POOL_COLD_KEY:
+		DENY_IF(violatesSingleAccountOrStoreIt(path));
 		WARN_UNLESS(bip44_isPathReasonable(path));
 		ALLOW();
+		break;
+
+	case PATH_POOL_COLD_KEY:
+		WARN_UNLESS(bip44_isPathReasonable(path));
+		SHOW();
 		break;
 
 	case PATH_MINT_KEY:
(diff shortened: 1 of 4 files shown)


src/signTx.h
```diff
@@ -49,12 +49,19 @@ enum {
 	SIGN_MAX_WITNESSES = SIGN_MAX_INPUTS + SIGN_MAX_OUTPUTS + SIGN_MAX_CERTIFICATES + SIGN_MAX_REWARD_WITHDRAWALS,
 };
 
+typedef struct {
+	bool pathSet;
+	bip44_path_t path;
```
janmazak, line 54: Isn't it enough to store just the account number?

src/singleAccount.c
```diff
@@ -0,0 +1,24 @@
+#include "state.h"
+
+static common_tx_data_t* commonData = &(instructionState.signTxContext.commonTxData);
+
+bool violatesSingleAccount(const bip44_path_t* path)
```
janmazak, line 5: As far as I understand, this both checks for violation and modifies ctx.

I am strongly in favor of separating those two things.

janmazak: I guess I'd like if ctx modifications were done in `signTx*.c`. (Is it true that we only do them there?)

src/signTxUtils.c
```diff
@@ -3,10 +3,39 @@
...
+static single_account_data_t* singleAccountData = &(instructionState.signTxContext.commonTxData.singleAccountData);
 
 void respondSuccessEmptyMsg()
 {
 	TRACE();
 	io_send_buf(SUCCESS, NULL, 0);
 	ui_displayBusy(); // needs to happen after I/O
 }
+
+bool violatesAndStoreSingleAccount(const bip44_path_t* path)
```
janmazak, line 18: Let's add a comment describing this function.

src/signTxUtils.c
```diff
@@ -3,10 +3,39 @@
 #include "uiHelpers.h"
 #include "utils.h"
 #include "signTxUtils.h"
+#include "securityPolicy.h"
+#include "state.h"
+
+static single_account_data_t* singleAccountData = &(instructionState.signTxContext.commonTxData.singleAccountData);
```
janmazak, line 9: It seems this is only needed in a single place, let's move it to `violatesAndStoreSingleAccount`.

src/signTxUtils.h
```diff
@@ -1,6 +1,11 @@
 #ifndef H_CARDANO_APP_SIGN_TX_UTILS
 #define H_CARDANO_APP_SIGN_TX_UTILS
 
+#include "signTx.h"
```
janmazak, line 4: Is this needed?

src/signTxUtils.c
janmazak:
```suggestion
bool violatesSingleAccountOrStoreIt(const bip44_path_t* path)
```

src/securityPolicy.c
```diff
@@ -703,6 +708,8 @@ static inline security_policy_t _ordinaryWitnessPolicy(const bip44_path_t* path,
 {
 	switch (bip44_classifyPath(path)) {
 	case PATH_ORDINARY_SPENDING_KEY:
+		DENY_IF(violatesAndStoreSingleAccount(path));
+	// intentional fallthrough
```
janmazak, line 712: Let's not use any fallthrough here, security policies is not a place for anything error-prone.

janmazak: Is there any reason to *not* check for single account violation in case of PATH_ORDINARY_STAKING_KEY?

src/signTxUtils.c
```diff
@@ -3,10 +3,39 @@
...
+	}
+	const bool byron = bip44_hasByronPrefix(path);
+	const uint32_t account = bip44_getAccount(path);
+	if (singleAccountData->pathSet) {
+		const uint32_t storedAccount = singleAccountData->accountNumber;
+		if (account != storedAccount) {
+			return true;
+		}
+		if ((singleAccountData->byron ^ byron)
+		    && storedAccount != 0) {
```
janmazak, line 32:
```suggestion
		bool combinesByronAndShelley = singleAccountData->byron ^ byron;
		bool combinationAllowed = (storedAccount == 0);
		if (combinesByronAndShelley && !combinationAllowed) {
```

janmazak (changes requested):

src/signTx.h
```diff
@@ -49,12 +49,20 @@ enum {
 	SIGN_MAX_WITNESSES = SIGN_MAX_INPUTS + SIGN_MAX_OUTPUTS + SIGN_MAX_CERTIFICATES + SIGN_MAX_REWARD_WITHDRAWALS,
 };
 
+typedef struct {
```
gabrielKerekes, line 52: I'd rename the fields here.
`pathSet` -> `isPathSet`
`byron` -> `isByron`
Similarly in `violatesAndStoreSingleAccount` then.

src/signTxUtils.c
gabrielKerekes: Wouldn't it more readable to use `singleAccountData->byron != byron` instead of `singleAccountData->byron ^ byron`? If I am not mistaken the result should be the same.

src/securityPolicy.c
gabrielKerekes: IMHO staking paths should also be checked. They are being checked on Trezor.

src/signTxUtils.c
```diff
@@ -3,10 +3,40 @@
...
+		ASSERT(false);
+	}
+	const bool byron = bip44_hasByronPrefix(path);
+	const uint32_t account = bip44_getAccount(path);
+	if (singleAccountData->pathSet) {
+		const uint32_t storedAccount = singleAccountData->accountNumber;
+		if (account != storedAccount) {
+			return true;
+		}
+		const bool combinesByronAndShelley = singleAccountData->byron ^ byron;
```
gabrielKerekes, line 31: Wouldn't `singleAccountData->byron != byron;` be a more standard way of checking unequality? I am not mistaken the result would be the same?

gabrielKerekes (approved):
Just minor comments from me since I was already looking at this. Otherwise, just by reading this, this seems to be consistent with what Trezor is doing. This definitely deserves some tests in LedgerJS though. [Below this line](https://github.com/trezor/trezor-firmware/blob/master/common/tests/fixtures/cardano/sign_tx.failed.json#L1585) are the Trezor tests just for comparison.

src/signTxUtils.c
mkv-vcm: I feel that the xor operator expresses the intent more clearly: either this or that is true but not both. Inequality is better when there are more than two options for values, no?

src/signTx.h
mkv-vcm: The conventions are more akin to my version I think, like `bool auxDataReceived`. I'm on the fence with isByron.

src/signTxUtils.c
janmazak: Well, xor evokes some vague fear of hard-to-understand bitwise magic in me :D 
So I find inequality easier and faster to comprehend.

janmazak (approved):

src/signTxUtils.c
mkv-vcm: Easy fix.

PR #126, opened 2021-10-25
Validate owner witness

mkv-vcm:
Remade PR to continue discussion.

### DIFF
--- a/src/securityPolicy.c
+++ b/src/securityPolicy.c
@@ -550,6 +550,35 @@ security_policy_t policyForSignTxCertificateStakePoolRetirement(
 	DENY(); // should not be reached
 }
 
+security_policy_t policyForSignTxStakePoolRegistrationInit(
+        sign_tx_signingmode_t txSigningMode,
+        size_t numOwners
+)
+{
+	switch (txSigningMode) {
+	case SIGN_TX_SIGNINGMODE_POOL_REGISTRATION_OWNER:
+		// there should be exactly one owner given by path for which we provide a witness
+		DENY_IF(numOwners == 0);
+		ALLOW();
+		break;
+
+	case SIGN_TX_SIGNINGMODE_POOL_REGISTRATION_OPERATOR:
+		ALLOW();
+		break;
+
+	case SIGN_TX_SIGNINGMODE_ORDINARY_TX:
+	case SIGN_TX_SIGNINGMODE_MULTISIG_TX:
+		DENY();
+		break;
+
+	default:
+		ASSERT(false);
+		break;
+	}
+
+	DENY(); // should not be reached
+}
+
 security_policy_t policyForSignTxStakePoolRegistrationPoolId(
         sign_tx_signingmode_t txSigningMode,
         const pool_id_t* poolId
@@ -613,7 +642,8 @@ security_policy_t policyForSignTxStakePoolRegistrationRewardAccount(
 
 security_policy_t policyForSignTxStakePoolRegistrationOwner(
         const sign_tx_signingmode_t txSigningMode,
-        const pool_owner_t* owner
+        const pool_owner_t* owner,
+        uint16_t numOwnersGivenByPath
 )
 {
 	if (owner->keyReferenceType == KEY_REFERENCE_PATH) {
@@ -623,10 +653,14 @@ security_policy_t policyForSignTxStakePoolRegistrationOwner(
 
 	switch (txSigningMode) {
 	case SIGN_TX_SIGNINGMODE_POOL_REGISTRATION_OWNER:
+		// can be 0 while processing owners given by hash
+		// or if no path owner is given at all (then we just compute the tx hash and don't allow witnesses)
+		DENY_UNLESS(numOwnersGivenByPath <= 1);
 		SHOW();
 		break;
 
 	case SIGN_TX_SIGNINGMODE_POOL_REGISTRATION_OPERATOR:
+		ASSERT(numOwnersGivenByPath == 0);
 		DENY_UNLESS(owner->keyReferenceType == KEY_REFERENCE_HASH);
 		SHOW();
 		break;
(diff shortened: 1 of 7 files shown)


src/securityPolicy.c
```diff
@@ -547,6 +547,35 @@ security_policy_t policyForSignTxCertificateStakePoolRetirement(
...
+	case SIGN_TX_SIGNINGMODE_MULTISIG_TX:
+		DENY();
+		break;
+
+	default:
+		ASSERT(false);
+		break;
+	}
+
+	DENY();
```
janmazak, line 576:
```suggestion
	DENY(); // should not be reached
```

src/securityPolicy.c
```diff
@@ -619,10 +649,14 @@ security_policy_t policyForSignTxStakePoolRegistrationOwner(
 
 	switch (txSigningMode) {
 	case SIGN_TX_SIGNINGMODE_POOL_REGISTRATION_OWNER:
+		// can be 0 because multiple owners can be given by third party
+		// or no path given at all, which is a pointless, but valid tx
```
janmazak, line 653:
```suggestion
		// can be 0 while processing owners given by hash
		// or if no path owner is given at all (then we just compute the tx hash and don't allow witnesses)
```

src/securityPolicy.c
```diff
@@ -739,12 +773,15 @@ static inline security_policy_t _scriptWitnessPolicy(const bip44_path_t* path, b
...
+static inline security_policy_t _poolRegistrationOwnerWitnessPolicy(const bip44_path_t* witnessPath, const bip44_path_t* poolOwnerPath)
 {
-	switch (bip44_classifyPath(path)) {
+	switch (bip44_classifyPath(witnessPath)) {
 
 	case PATH_ORDINARY_STAKING_KEY:
-		WARN_UNLESS(bip44_isPathReasonable(path));
+		if (poolOwnerPath != NULL) {
+			DENY_UNLESS(bip44_pathsEqual(witnessPath, poolOwnerPath));
+		}
```
janmazak, line 783:
```suggestion
		if (poolOwnerPath != NULL) {
			// an owner was given by path
			// the witness path must be identical
			DENY_UNLESS(bip44_pathsEqual(witnessPath, poolOwnerPath));
		} else {
			// no owner was given by path
			// we must not allow witnesses because they might witness owners given by key hash
			DENY();		
		}
```

janmazak: We also need a ledgerjs test for this case.

src/securityPolicy.h
```diff
@@ -84,7 +88,8 @@ security_policy_t policyForSignTxStakePoolRegistrationRewardAccount(
 );
 security_policy_t policyForSignTxStakePoolRegistrationOwner(
         const sign_tx_signingmode_t txSigningMode,
-        const pool_owner_t* owner
+        const pool_owner_t* owner,
+        uint16_t numOwnersGivenByPath
```
janmazak, line 92: I think we are using `size_t` in other policies for such numbers.

src/signTxPoolRegistration.c
```diff
@@ -806,8 +791,8 @@ static void signTxPoolRegistration_handleOwnerAPDU(uint8_t* wireDataBuffer, size
 			PRINTF("\n");
 
 			subctx->numOwnersGivenByPath++;
-			VALIDATE(subctx->numOwnersGivenByPath <= 1, ERR_INVALID_DATA);
-
+			ctx->poolOwnerByPath = true;
+			memcpy(&ctx->poolOwnerPath, &owner->path, SIZEOF(owner->path));
```
janmazak, line 795:
```suggestion
			memmove(&ctx->poolOwnerPath, &owner->path, SIZEOF(owner->path));
```

janmazak: This storing of the owner path in ctx should not be done here, only after the security policy has been checked. And we should add an assert before storing that checks that nothing was stored before.

janmazak (changes requested):
There are some small but important changes related to security.

src/signTxPoolRegistration.c
mkv-vcm: We write a lot of stuff into the context, why this is the one you think should happen after policy?

src/securityPolicy.h
mkv-vcm: I don't think so, there are plenty of examples where we use the original data type of the data.

src/securityPolicy.c
mkv-vcm: Done.

src/securityPolicy.c
mkv-vcm: Done, but can't merge until it's resolved.

src/securityPolicy.h
janmazak: You are right, I was not checking carefully. There are two places where we use size_t:
```
security_policy_t policyForSignTxStakePoolRegistrationInit(
        sign_tx_signingmode_t txSigningMode,
        size_t numOwners
)
```
and in
```
security_policy_t policyForGetPublicKeysInit(size_t numPaths)
```
And some where we use `uint32_t`:
```
security_policy_t policyForSignTxStakePoolRegistrationConfirm(
        uint32_t numOwners, uint32_t numRelays
);
```

Considering that the numbers are serialized as uint32_t in APDUs and we only use uint16 to save memory in ctx, I am in favor of using uint32_t in all the security policies. Would you change it, please? Maybe in a separate PR?

src/signTxPoolRegistration.c
janmazak: You are right, we do VALIDATE after storing stuff in ctx, too. No need to change it then.

But the ASSERT would be reasonable (since we never want to replace an already stored value), woudn't it?

janmazak (approved):
`make format` should be run.

src/signTxPoolRegistration.c
mkv-vcm: The Assert is already there, not sure what version you are seeing.

PR #127, opened 2021-10-25
name view handler functions more consistently

mkv-vcm:
Also renamed the appendData function, because it wasn't consistent with the read version.


janmazak (approved):
After you fix the stuff in the only comment, squash it into a single commit and merge it into develop.

PR #128, opened 2021-11-03
Add security policy comments

davidmisiak:


src/securityPolicy.c
```diff
@@ -69,22 +69,23 @@ security_policy_t policyForDerivePrivateKey(const bip44_path_t* path)
...
-
 		ALLOW();
+		break;
 
 	default:
 		DENY();
+		break;
 	}
+
+	DENY(); // should not be reached
```
davidmisiak, line 85: I know that editing this function was not necessary, I just wanted to make it visually aligned with `policyForGetExtendedPublicKey` and `policyForGetExtendedPublicKeyBulkExport` below.

src/securityPolicy.c
```diff
@@ -717,6 +768,8 @@ security_policy_t policyForSignTxWithdrawal(
 		break;
```
davidmisiak, line 768: TODO add a comment here when Plutus withdrawal policy is determined

src/securityPolicy.c
```diff
@@ -806,6 +874,7 @@ static inline security_policy_t _poolRegistrationOperatorWitnessPolicy(const bip
 
 	case PATH_ORDINARY_SPENDING_KEY:
 	case PATH_POOL_COLD_KEY:
+		// only ordinary spending key paths (because of inputs) and pool cold key path are allowed
 		WARN_UNLESS(bip44_isPathReasonable(path));
 		SHOW();
```
davidmisiak, line 879: Why to we need to show ordinary key paths here?

src/securityPolicy.c
```diff
@@ -884,6 +957,7 @@ security_policy_t policyForSignTxMintConfirm(security_policy_t outputPolicy)
 		break;
```
davidmisiak, line 957: If I'm not mistaken, the `outputPolicy` variable here is actually mint policy from `policyForSignTxMintInit`, so we should
1. rename it,
2. remove the `POLICY_ALLOW_WITHOUT_PROMPT` case here, since it's unreachable,

right?

janmazak: I've renamed it, but I don't want to remove this currently unreachable switch case (it's OK to have this policy/UI machine ready, does not cost us a lot of code).

src/securityPolicy.c
janmazak: I don't see any reason for that, but am wary of not showing them, I've just added a TODO.

PR #131, opened 2021-11-18
fix a single account security model bug

janmazak:
Why didn't our ledgerjs tests catch this?


mkv-vcm:
> Why didn't our ledgerjs tests catch this?

Because it's on the positive side, not the negative, and I only made rejection tests, no Byron-Shelley positive test was created.

mkv-vcm (approved):

PR #133, opened 2021-11-24
refactor tx init policy

janmazak:


src/securityPolicy.c
```diff
@@ -212,6 +212,31 @@ security_policy_t policyForShowDeriveAddress(const addressParams_t* addressParam
...
+	        networkId == MAINNET_NETWORK_ID ||
+	        networkId == TESTNET_NETWORK_ID;
+	const bool usualProtocolMagic = protocolMagic == MAINNET_PROTOCOL_MAGIC;
+	return usualNetworkId && usualProtocolMagic;
+}
+
+bool needsMissingCollateralWarning(sign_tx_signingmode_t signingMode, uint32_t numCollaterals)
+{
+	const bool collateralExpected = signingMode == SIGN_TX_SIGNINGMODE_PLUTUS_TX;
+	return collateralExpected && (numCollaterals == 0);
```
mkv-vcm, line 227: collateral**s**, to keep the naming consistent.

src/securityPolicy.c
```diff
@@ -256,17 +281,20 @@ security_policy_t policyForSignTxInit(
...
 
-	WARN_IF(networkId != MAINNET_NETWORK_ID && networkId != TESTNET_NETWORK_ID);
-	WARN_IF(protocolMagic != MAINNET_PROTOCOL_MAGIC);
+	// there are separate screens for various warnings
+	// the return value of the policy only says that at least one should be applied
+	// and the need for individual warnings is reassessed in the UI machine
+	WARN_UNLESS(isNetworkUsual(networkId, protocolMagic));
+	WARN_IF(needsRunningScriptWarning(numCollaterals));
+	WARN_IF(needsMissingCollateralWarning(txSigningMode, numCollaterals));
+	WARN_IF(needsMissingScriptDataHashWarning(txSigningMode, includeScriptDataHash));
```
mkv-vcm, line 297: Not sure I like these two outside of the signing mode switch, since they are obviously related to that one signing mode.

src/securityPolicy.h
```diff
@@ -18,6 +18,10 @@ security_policy_t policyForGetExtendedPublicKeyBulkExport(const bip44_path_t* pa
 security_policy_t policyForShowDeriveAddress(const addressParams_t* addressParams);
 security_policy_t policyForReturnDeriveAddress(const addressParams_t* addressParams);
 
+bool isNetworkUsual(uint32_t networkId, uint32_t protocolMagic);
+bool needsRunningScriptWarning(int32_t numCollaterals);
```
mkv-vcm, line 22: I would prefer if functions in the header would be in the same order as they are in the .c file.

janmazak: That was my intent; it is hopefully fixed now.

src/securityPolicy.c
janmazak: I thought about that quite a lot, but I don't think it is obvious from their name or their use in `signTx.c` that it is only relevant for a single signing mode. It is true for now, but we might later find out that we need more Plutus tx modes or whatever, and keeping them here will withstand such a change better.

Also, the UI state machine displaying warnings does not distinguish between signing modes, and I want to keep the logic deciding which warnings to show as much the same as I can make it.

src/securityPolicy.c
janmazak: I view it more generally --- collateral inputs are just a particular way of providing collateral for Plutus scripts (the only one existing at present). Calling collateral inputs "collaterals" is useful to keep variable names short in our code, but I don't particularly like it, it's not called like that anywhere else, so I tend to avoid it if I can.

janmazak:
I've not run any tests on this branch.

mkv-vcm:
Tested it, seems to be working as intended.

PR #134, opened 2021-11-26
update Makefile

janmazak:
@mkv: Please test if this works. Also, notice new make targets (esp. analyze --- running clang this way is part of CI).
It's one of the refactors we want to bundle with Plutus.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,40 +15,50 @@
 #  limitations under the License.
 #*******************************************************************************
 
-ifeq ($(BOLOS_SDK),)
-$(error Environment variable BOLOS_SDK is not set)
-endif
-include $(BOLOS_SDK)/Makefile.defines
-
-SIGNKEY = `cat sign.key`
-APPSIG = `cat bin/app.sig`
-NANOS_ID = 1
-WORDS = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about"
-PIN = 5555
-
 APPNAME      = "Cardano ADA"
 APPVERSION_M = 3
 APPVERSION_N = 0
 APPVERSION_P = 0
 APPVERSION   = "$(APPVERSION_M).$(APPVERSION_N).$(APPVERSION_P)"
 
-APP_LOAD_PARAMS =--appFlags 0x240 --curve ed25519 --path "44'/1815'" --path "1852'/1815'" --path "1853'/1815'" --path "1854'/1815'" --path "1855'/1815'"
-APP_LOAD_PARAMS += $(COMMON_LOAD_PARAMS)
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -55,9 +55,25 @@ The build process is managed with [Make](https://www.gnu.org/software/make/).
...
+* the device is in some intermediary state (e.g. processing an APDU message)
+* the app is too big to fit on the device (try `make size`)
+
+### Before merging a pull request
+
+_Before merging a PR, one should make sure that:_
+* `make format` does not change anything (except possibly some glyph* files)
+* `make clean load` runs without errors and warnings (except those reported for nanos-secure-sdk repo) for production build
+* `make clean load` runs without errors and warnings (except those reported for nanos-secure-sdk repo) for development build (see Debug version above)
+* `make analyze` does not report errors or warnings
```
mkv-vcm, line 75: This will be tricky, I can't yet figure out how to blacklist the SDK folder or whitelist the project source. And the SDK has a problem with the snprintf (what else).

Makefile
```diff
@@ -169,19 +177,20 @@ seed:
...
+#    Size    #
 ##############
 
-# better to run this manually to avoid irrelevant dependencies processing
-format:
-	astyle --options=.astylerc src/*.h src/*.c
+# prints app size, max is about 140K
+
+size: all
+	$(GCCPATH)arm-none-eabi-size --format=gnu bin/app.elf
```
mkv-vcm, line 195: Interesting. With GNU size (GNU Tools for ARM Embedded Processors) 2.26.0.20160310, the --format does not accept gnu. --format={sysv|berkeley}

mkv-vcm (commented):
Tried some stuff, will try some more stuff on Monday, to make this change work, because right now it's quesitonable.

Makefile
janmazak: I'm using GNU size (GNU Arm Embedded Toolchain 10-2020-q4-major) 2.35.1.20201028 and there it works. Perhaps we can use --format=berkeley, it's about the same as gnu, but I like "total" better than "dec" as column name.

README.md
janmazak: I keep the SDK separated, but if you figure out how to make the command more specific to the project source, let's use it.

mkv-vcm: Ah yeah, it's my bad, I experimented with various SDKs (for the new, docker based build recommended by the ledger dev website), and one was left in the ledger directory.

Makefile
mkv-vcm: That's fine, I think we want to keep the up to date option.

mkv-vcm (approved):

PR #135, opened 2021-11-30
UI output correction for pool retirement

gitmachtl:
The UI is showing "... retirement at the **end** of epoch xxx" for confirmation. Actually the retirement is at the **begin** of the retirement epoch, not at the end. Text still fits into the Nano S display. word could also be changed into start. Just a minor fix.

### DIFF
--- a/src/signTx.c
+++ b/src/signTx.c
@@ -1045,7 +1045,7 @@ static void signTx_handleCertificatePoolRetirement_ui_runStep()
 	}
 	UI_STEP(HANDLE_CERTIFICATE_POOL_RETIREMENT_STEP_DISPLAY_EPOCH) {
 		ui_displayUint64Screen(
-		        "at the begin of epoch",
+		        "at the start of epoch",
 		        BODY_CTX->stageData.certificate.epoch,
 		        this_fn
 		);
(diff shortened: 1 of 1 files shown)


src/signTx.c
```diff
@@ -4,7 +4,7 @@
 #include "cardano.h"
 #include "addressUtilsByron.h"
 #include "addressUtilsShelley.h"
-#include "uiHelpers.h"
+#include "uiHelpers.h"e
```
gitmachtl: this was a typo, please ignore...

src/signTx.c
```diff
@@ -1045,7 +1045,7 @@ static void signTx_handleCertificatePoolRetirement_ui_runStep()
 	}
 	UI_STEP(HANDLE_CERTIFICATE_POOL_RETIREMENT_STEP_DISPLAY_EPOCH) {
 		ui_displayUint64Screen(
-		        "at the end of epoch",
+		        "at the begin of epoch",
```
janmazak, line 1048: I'd use "at beginning of epoch" or "at the start of epoch" --- what do you think, @gitmachtl?

janmazak (commented):
We are going to make the change, just the precise wording of the message needs to be determined.

src/signTx.c
gitmachtl: i would go with "at the start of epoch" than, because of the wordwrap. it fits into one line in that case.

janmazak (approved):

PR #137, opened 2021-12-05
change certain thrown exceptions

janmazak:
The code would be more consistent this way (we generally use ASSERT to detect our failures, not THROW exceptions that must not "go out").


src/errors.h
```diff
@@ -22,10 +22,11 @@ enum {
...
 	// Request is not valid in the context of previous calls
 	ERR_INVALID_STATE              = 0x6E06,
-	// Some part of request data is invalid (includes not enough data)
+	// Some part of request data is invalid (or unknown to this app)
+	// (includes not enough data and too much data)
 	ERR_INVALID_DATA               = 0x6E07,
-	// BIP44 path is rejected
-	ERR_INVALID_BIP44_PATH         = 0x6E08,
+	// previously used for rejected BIP44 paths, now we use ERR_REJECTED_BY_POLICY
+	// ERR_INVALID_BIP44_PATH         = 0x6E08,
```
mkv-vcm, line 29: From the comment I assume you intend to leave these in the code. What purpose do they serve?

mkv-vcm (approved):
Good direction.

src/errors.h
janmazak: I'd prefer not to reuse the error code (e.g. 0x6E08) in the future (to avoid potential confusion if someone has an enum for these codes etc.).

PR #138, opened 2021-12-05
format preprocessor directives more consistently

janmazak:
A trivial change.


mkv-vcm (approved):
Yes, we are not using OLD C, so this should improve readability without compromising anything.

PR #141, opened 2022-04-28
fix ugly text in UI

janmazak:


davidmisiak (approved):
I checked the code, signed a few txs manually (observing the modified messages) and ran Ledgerjs integration tests (just to make sure), everything looks good.

However, I noticed that there are some other strings that could be split into two lines like `Confirm output`, namely `Confirm mint`, `No pool owners` and `No pool relays`.

PR #144, opened 2022-05-11
Fix device tests

janmazak:


src/textUtils.c
```diff
@@ -12,7 +12,7 @@ size_t str_formatAdaAmount(uint64_t amount, char* out, size_t outSize)
 {
 	ASSERT(outSize < BUFFER_SIZE_PARANOIA);
 
-	char scratchBuffer[30] = {0};
```
rustamlabs, line 15: Is that change because of the changes in token decimals?

janmazak: I think yes. This function will be rewritten a bit when decimal places are added; I had all those changes bundled on a single branch and extracted some of them here so that we have a working device test. It might be that not exactly everything is necessary, but it is not worth thinking about this too much.

rustamlabs (approved):

PR #145, opened 2022-05-16
Item 17

rustamlabs:


src/signTx.c
```diff
@@ -234,6 +235,15 @@ static inline void advanceStage()
...
 		}
+		ctx->stage = SIGN_STAGE_BODY_TOTAL_COLLATERALS;
+		break;
+
+	case SIGN_STAGE_BODY_TOTAL_COLLATERALS:
+		if (ctx->includeTotalCollateral) {
+			// TODO: Any assertion here? (sum of ADA in collateral inputs) = txColl + (ADA in collRet)
+			txHashBuilder_enterTotalCollaterals(&BODY_CTX->txHashBuilder, ctx->txColl);
+		}
+		//TODO: should we assert for (sum of ADA in collateral inputs) >= (min collateral requirement) + (ADA in collRet) as the "else" case?
```
janmazak, line 246: Does not belong in here, see `securityPolicy.c` where such conditions are checked and then applied (resulting in UI warnig messages, tx rejection etc.)

src/txHashBuilder_test.c
```diff
@@ -374,7 +374,8 @@ void run_txHashBuilder_test()
 	                   false, // script hash data
 	                   0,	// collaterals not tested yet
 	                   0,	// required signers not tested yet
-	                   false // network id
+	                   false, // network id
+	                   true // total collaterals
```
janmazak, line 378:
```suggestion
	                   true // total collateral
```

src/signTx.h
```diff
@@ -138,6 +140,7 @@ typedef struct {
 	uint16_t currentWithdrawal;
 	uint16_t currentCollateral;
 	uint16_t currentRequiredSigner;
+	uint16_t currentReferenceInputs;
```
janmazak, line 143:
```suggestion
	uint16_t currentReferenceInput;
```

src/txHashBuilder.c
```diff
@@ -54,6 +54,76 @@ static void blake2b_256_append_cbor_tx_body(
...
+		BUILDER_APPEND_CBOR(CBOR_TYPE_UNSIGNED, amount);
+	}
+
+	//TODO: Move this to approperiate place. Where is the appropriate place? New method for adding post alonzo txout? or the existing one?
+	//or does advancing the stae in cbor_append methods makes sense?
+	if (includeDatumOption) {
+		builder->state = TX_HASH_BUILDER_IN_OUTPUTS_DATUM_OPTION;
+	}
+	if (includeScriptRef) {
+		builder->state = TX_HASH_BUILDER_IN_OUTPUTS_SCRIPT_REFERANCE;
```
janmazak, line 120:
```suggestion
		builder->state = TX_HASH_BUILDER_IN_OUTPUTS_SCRIPT_REFERENCE;
```

src/txHashBuilder.c
```diff
@@ -127,7 +203,16 @@ void txHashBuilder_init(
...
 
-		ASSERT((3 <= numItems) && (numItems <= 13));
+		builder->includeCollateralReturn = includeCollateralReturn;
+		if (includeCollateralReturn) numItems++;
+
+		builder->includeTotalCollateral = includeTotalCollateral;
+		if (includeTotalCollateral) numItems++;
+
+		builder->remainingReferenceInputs = numReferenceInputs;
+		numItems++; // an array that is always included (even if empty)
```
janmazak, line 213: No, take a look at https://cips.cardano.org/cips/cip21/:
"Unless mentioned otherwise in this CIP, optional empty lists and maps must not be included as part of the transaction body or its elements."

We don't allow empty arrays for optional elements (cf. `numCollaterals`).

src/txHashBuilder.c
```diff
@@ -387,6 +447,38 @@ void txHashBuilder_addOutput_datumHash(
 	builder->state = TX_HASH_BUILDER_IN_OUTPUTS;
 }
 
+void txHashBuilder_addOutput_datumOption(
+        tx_hash_builder_t* builder,
+        const uint8_t* bytesBuffer, size_t bufferSize, uint8_t datumOption, bool includeScriptref
```
janmazak, line 452: We want an enum for better readability for `datumOption`, and I'd pass it before the buffer. Also, rename bytesBuffer -> buffer for consistency.

janmazak: Passing `includeScriptref` as an argument for the sole purpose of changing the internal txhashbuilder state smells like a bad design: it is not related to datumOption.

Is this approach copied from somewhere? Perhaps we should rethink the internal builder state. I'd expect only `cbor_append_post_alonzo_txOutput` to receive `includeScriptref`. Maybe we don't have a place to store the information in `tx_hash_builder_t`?

src/txHashBuilder.c
```diff
@@ -1466,9 +1545,168 @@ static void txHashBuilder_assertCanLeaveNetworkId(tx_hash_builder_t* builder)
...
-void txHashBuilder_finalize(tx_hash_builder_t* builder, uint8_t* outBuffer, size_t outSize)
+void txHashBuilder_enterCollateralReturn(tx_hash_builder_t *builder)
 {
+	_TRACE("state = %d", builder->state);
+
 	txHashBuilder_assertCanLeaveNetworkId(builder);
+	{
+		// Enter collateral output
+		BUILDER_APPEND_CBOR(CBOR_TYPE_UNSIGNED, TX_BODY_KEY_COLLATERAL_RETURN);
+		BUILDER_APPEND_CBOR(CBOR_TYPE_ARRAY, builder->includeCollateralReturn);
```
janmazak, line 1556: This looks suspicious, why an array? Item 16 is a single element of type `transaction_output = legacy_transaction_output / post_alonzo_transaction_output`, so two possibilities.

I don't think we even need `txHashBuilder_enterCollateralReturn`, those `txHashBuilder_enter*` functions are for arrays/maps.

src/txHashBuilder.c
```diff
@@ -1466,9 +1545,168 @@ static void txHashBuilder_assertCanLeaveNetworkId(tx_hash_builder_t* builder)
...
+	ASSERT(builder->remainingReferenceInputs > 0);
+	{
+		// Enter reference inputs
+		BUILDER_APPEND_CBOR(CBOR_TYPE_UNSIGNED, TX_BODY_KEY_REFERENCE_INPUTS);
+		BUILDER_APPEND_CBOR(CBOR_TYPE_ARRAY, builder->remainingReferenceInputs);
+	}
+	builder->state = TX_HASH_BUILDER_IN_REFERENCE_INPUTS;
+
+}
+void txHashBuilder_addReferenceInputs(
```
janmazak, line 1661:
```suggestion
void txHashBuilder_addReferenceInput(
```

src/txHashBuilder.c
```diff
@@ -1466,9 +1545,168 @@ static void txHashBuilder_assertCanLeaveNetworkId(tx_hash_builder_t* builder)
...
+        tx_hash_builder_t* builder,
+        const uint8_t* utxoHashBuffer, size_t utxoHashSize,
```
janmazak, line 1663: Since we use inputs in various places, it might make sense to create a structure instead of passing a triple of arguments all around. We actually already do have the type: `sign_tx_transaction_input_t`. It just needs to be moved to `cardano.h` and renamed to, say, `tx_input_t`.

src/txHashBuilder.c
```diff
@@ -1466,9 +1545,168 @@ static void txHashBuilder_assertCanLeaveNetworkId(tx_hash_builder_t* builder)
...
+	case TX_HASH_BUILDER_IN_REQUIRED_SIGNERS:
+	case TX_HASH_BUILDER_IN_COLLATERALS:
+	case TX_HASH_BUILDER_IN_SCRIPT_DATA_HASH:
+	case TX_HASH_BUILDER_IN_MINT:
+	case TX_HASH_BUILDER_IN_VALIDITY_INTERVAL_START:
+	case TX_HASH_BUILDER_IN_AUX_DATA:
+	case TX_HASH_BUILDER_IN_WITHDRAWALS:
+	case TX_HASH_BUILDER_IN_CERTIFICATES:
+	case TX_HASH_BUILDER_IN_TTL:
+	case TX_HASH_BUILDER_IN_FEE:
```
janmazak, line 1697: This is becoming annoying. Give it a while and try to come up with some reasonable solution to avoid these long lists in `txHashBuilder_assertCanLeave*` functions (we should discuss it before coding it). Perhaps using comparisons between states?... Inequalities are implicit and so more error-prone vs. explicit lists (e.g. when we add states), but these lists are becoming too long to read/check, and we don't expect to put states in between the existing ones.

src/txHashBuilder.c
```diff
@@ -54,6 +54,138 @@ static void blake2b_256_append_cbor_tx_body(
...
+		}
+		{
+			BUILDER_APPEND_CBOR(CBOR_TYPE_UNSIGNED, output->amount);
+		}
+		builder->state = TX_HASH_BUILDER_IN_OUTPUTS_TOP_LEVEL_DATA;
+	} else {
+		builder->multiassetData.remainingAssetGroups = output->numAssetGroups;
+		// Array(2)[
+		//   Bytes[address]
+		//   Array(2)[]
```
janmazak, line 98: It seems this does not allow to add datum hash when tokens are present. That would be incorrect.

src/txHashBuilder.c
```diff
@@ -54,6 +54,138 @@ static void blake2b_256_append_cbor_tx_body(
...
+			BUILDER_APPEND_CBOR(CBOR_TYPE_UNSIGNED, output->amount);
+		}
+
+		builder->state = TX_HASH_BUILDER_IN_OUTPUTS_TOP_LEVEL_DATA;
+	} else {
+		builder->multiassetData.remainingAssetGroups = output->numAssetGroups;
+		// Map(2 + includeDatumOption + includeScriptRef)[
+		//   Bytes[address]
+		//   Array(2)[]
+		//     Unsigned[amount]
```
janmazak, line 151: What about tokens? We need to support them in the new format, too.

src/signTxOutput.h
```diff
@@ -51,6 +53,7 @@ typedef struct {
 	uint16_t currentToken;
 	bool includeDatumHash;
```
janmazak, line 54: This should be updated: `output_context_t` must support everything needed, including the post alonzo stuff. But first let's get txhashbuilder finished.

src/txHashBuilder.c
```diff
@@ -368,23 +448,77 @@ void txHashBuilder_addOutput_token(
...
+	}
+	{
+		BUILDER_APPEND_CBOR(CBOR_TYPE_BYTES, bufferSize);
+		BUILDER_APPEND_DATA(buffer, bufferSize);
+	}
+	builder->state = TX_HASH_BUILDER_IN_OUTPUTS_DATUM_OPTION;
+}
+
+
+void txHashBuilder_addOutput_referenceScript(tx_hash_builder_t *builder, const uint8_t *scriptBuffer, size_t bufferSize)
```
janmazak, line 510: We need functions that will allow to add chunks of the script repeatedly. So the first function will only add the tag (6.24) and total bytestring length (the tx hash builder state machine should make sure it can only be called once) and then another function will add bytes, and can be called repeatedly.

janmazak: An analogous pair of fucntions for inline datum.

src/txHashBuilder.h
```diff
@@ -91,9 +115,27 @@ typedef struct {
 	};
 
 	tx_hash_builder_state_t state;
+	tx_hash_builder_output_substate_t subState;
 	blake2b_256_context_t txHash;
 } tx_hash_builder_t;
 
+typedef enum  {
+	LEGACY = 0,
+	POST_ALONZO = 1
```
janmazak, line 124: As agreed elsewhere, we prefer to use names that are more perpetual (won't become invalid after the next serialization format change):
ARRAY_LEGACY and MAP_BABBAGE. You can add the CDDL names as comments to the enum values.

src/txHashBuilder.h
```diff
@@ -91,9 +115,27 @@ typedef struct {
 	};
 
 	tx_hash_builder_state_t state;
+	tx_hash_builder_output_substate_t subState;
```
janmazak, line 118: This smells of bad design --- a variable named "subState" should be generic, not related to just outputs. We can rename it (e.g. outputSubstate, and perhaps put it in an unnamed union as the one above, for now the union will have just a single member), but we need to think about how it is related to the union above --- maybe some of its items actually belong to the substate? I guess we don't want to overcomplicate this, but also don't want to have confusing names/structures.

src/txHashBuilder.h
```diff
@@ -54,29 +59,48 @@ typedef enum {
...
 	uint16_t remainingOutputs;
 	uint16_t remainingWithdrawals;
 	uint16_t remainingCertificates;
 	uint16_t remainingCollaterals;
 	uint16_t remainingRequiredSigners;
+	uint16_t remainingReferenceInputs;
+	uint16_t totalDatumSize;
+	uint16_t currentDatumSize;
+	uint16_t totalReferenceScriptSize;
+	uint16_t currentReferenceScriptSize;
```
janmazak, line 95: I'd move these elsewhere, they are not direct tx body elements (as the rest of this list). Perhaps the union below?

src/txHashBuilder.h
rustamlabs: In sign tx  `sign_tx_output_state_t` is used, maybe I can get rid of "sub" and make it as :
tx_hash_builder_output_state_t outputState?

janmazak: I guess yes, I'm not sure what you refer to.

PR #147, opened 2022-06-14
fixes for v4.1.0 (expert mode, token decimals)

janmazak:
This PR is to gather changes and bugfixes related to tokens / expert mode. When we finish gathering feedback, I'll create a PR for Ledger.


janmazak:
This PR is to gather changes and bugfixes related to tokens / expert mode. When we finish gathering feedback, I'll create a PR for Ledger.

davidmisiak (commented):
LGTM. I haven't run anything (let me know if I should test the changes).

Just to make sure - datum hash in non-script outputs is not intended to be allowed here, right? (I suppose this will be a part of Babbage changes in that case.)

PR #150, opened 2022-06-16
UI inputs hashbuilder

rustamlabs:


src/signTx.c
```diff
@@ -806,8 +806,8 @@ static void parseInput(uint8_t* wireDataBuffer, size_t wireDataSize)
 
 	VALIDATE(wireDataSize == SIZEOF(*wireUtxo), ERR_INVALID_DATA);
 
-	memmove(input->txHashBuffer, wireUtxo->txHash, SIZEOF(input->txHashBuffer));
-	input->parsedIndex = u4be_read(wireUtxo->index);
+	memmove(input->input_data.txHashBuffer, wireUtxo->txHash, SIZEOF(input->input_data.txHashBuffer));
```
janmazak, line 809: Instead of `sign_tx_transaction_input_t* input`, let's make a pointer to input->input_data at the beginning of this function.

janmazak: We can also move the pointer declaration closer to its use (after VALIDATE).

src/txHashBuilder.c
```diff
@@ -155,13 +155,10 @@ void txHashBuilder_enterInputs(tx_hash_builder_t* builder)
 	builder->state = TX_HASH_BUILDER_IN_INPUTS;
 }
 
-void txHashBuilder_addInput(
-        tx_hash_builder_t* builder,
-        const uint8_t* utxoHashBuffer, size_t utxoHashSize,
-        uint32_t utxoIndex
-)
+void txHashBuilder_addInput(tx_hash_builder_t *builder, tx_input_t input)
```
janmazak, line 158: We are generally passing arguments by pointers in such situations, I don't see a reason to copy the value here (it might even consume stack memory).

src/txHashBuilder.c
```diff
@@ -155,13 +155,10 @@ void txHashBuilder_enterInputs(tx_hash_builder_t* builder)
...
 {
 	_TRACE("state = %d, remainingInputs = %u", builder->state, builder->remainingInputs);
+	size_t utxoHashSize = SIZEOF(input.txHashBuffer);
```
janmazak, line 161: Let's use `const size_t utxoHashSize`.

src/txHashBuilder.c
```diff
@@ -1303,13 +1300,10 @@ void txHashBuilder_enterCollaterals(tx_hash_builder_t* builder)
 	builder->state = TX_HASH_BUILDER_IN_COLLATERALS;
 }
 
-void txHashBuilder_addCollateral(
-        tx_hash_builder_t* builder,
-        const uint8_t* utxoHashBuffer, size_t utxoHashSize,
-        uint32_t utxoIndex
-)
+void txHashBuilder_addCollateral(tx_hash_builder_t *builder, tx_input_t collInput)
```
janmazak, line 1303: A pointer here, too.

janmazak: It seems we would want to extract the common code that serializes `tx_input_t` into the tx. (I.e. there would be separate functions for adding inputs and collateral inputs that will handle the state machine logic, but serialization itself can be done by a common static function called from those two.)

src/uiScreens.c
```diff
@@ -823,22 +823,22 @@ void ui_displayInputScreen(
         const sign_tx_transaction_input_t* input,
         ui_callback_fn_t callback)
 {
-	char txHex[2 * SIZEOF(input->txHashBuffer) + 1] = {0};
+	char txHex[2 * SIZEOF(input->input_data.txHashBuffer) + 1] = {0};
```
janmazak, line 826: I'm not entirely sure, but maybe we can use some temporary variables here to shorten expression like `input->input_data.txHashBuffer`.

Perhaps assert that `SIZEOF(input->input_data.txHashBuffer) == TX_HASH_LENGTH` and only use TX_HASH_LENGTH afterwards for the buffer length; and perhaps `const uint8_t* hash = input->input_data.txHashBuffer`.

janmazak:
Let's remove the "draft" status from this PR. After it is approved, we want to squash all the commits into a single one and merge it into develop.

src/uiScreens.c
rustamlabs: I like the TX_HASH_LENGTH idea, but for hash I think it is better to have one pointer that points to input->input_data and use it 2 times than having 2 pointers (1 for hash 1 for index) and using them once or or having pointer for one and not for the other

janmazak (approved):

PR #165, opened 2022-07-22
small improvements for tx hash builder

janmazak:


src/txHashBuilder.c
```diff
@@ -349,14 +353,14 @@ void txHashBuilder_addOutput_topLevelData(
...
-                          tx_hash_builder_state_t mandatoryState)
+static void addTokenGroup(
+	tx_hash_builder_t* builder,
+	const uint8_t* policyIdBuffer, size_t policyIdSize,
+	uint16_t numTokens
+)
 {
 	_TRACE("state = %d, remainingAssetGroups = %u", builder->state, builder->multiassetData.remainingAssetGroups);
 
-	ASSERT(builder->state == mandatoryState); //TX_HASH_BUILDER_IN_OUTPUTS | TX_HASH_BUILDER_IN_MINT | TX_HASH_BUILDER_IN_COLLATERAL_RETURN
```
janmazak, line 359: I've removed these because they are local functions that are called only from functions that assert the state anyway.

rustamlabs: I thought that was a extra safety measure to check state on every call. What makes it unnecessary in this case?

janmazak: The check was really redundant: essentially, the functions calling this asserted that state is X, and then explicitly told this function to assert that the state is X. It is extremely unlikely that anyone will call a token-adding function in a wrong place (and that will not be noticed in code review nor in integration tests).

PR #176, opened 2022-10-19
governance (catalyst update and fragment signing)

janmazak:


src/bip44.c
```diff
@@ -118,6 +118,17 @@ bool bip44_hasPoolColdKeyPrefix(const bip44_path_t* pathSpec)
 #undef CHECK
 }
 
+// /1694'/1815'
+bool bip44_hasGovernanceVotingKeyPrefix(const bip44_path_t* pathSpec)
+{
+#define CHECK(cond) if (!(cond)) return false
+	CHECK(pathSpec->length > BIP44_I_COIN_TYPE);
+	CHECK(pathSpec->path[BIP44_I_PURPOSE] == (PURPOSE_GOVERNANCE_VOTING_KEY | HARDENED_BIP32));
+	CHECK(pathSpec->path[BIP44_I_COIN_TYPE] == (ADA_COIN_TYPE | HARDENED_BIP32));
```
davidmisiak, line 127: Maybe you could use the `harden` and `isHardened` functions here and in `bip44_isGovernanceVotingKeyPath`?

src/securityPolicy.c
```diff
@@ -162,6 +181,8 @@ security_policy_t policyForGetExtendedPublicKeyBulkExport(const bip44_path_t* pa
 	case PATH_MULTISIG_SPENDING_KEY:
 	case PATH_MULTISIG_STAKING_KEY:
 	case PATH_MINT_KEY:
+	case PATH_GOVERNANCE_VOTING_ACCOUNT:
+	case PATH_GOVERNANCE_VOTING_KEY:
```
davidmisiak, line 185: Unrelated to this specific line, but maybe we could add a governance key case also to `ui_displayGetPublicKeyPathScreen`?

src/bip44.c
```diff
@@ -388,6 +410,37 @@ static bip44_path_type_t bip44_classifyMultisigWalletPath(const bip44_path_t* pa
...
+	switch (pathSpec->length) {
+	case 3: {
+		return PATH_GOVERNANCE_VOTING_ACCOUNT;
+	}
+	case 5: {
+		if (bip44_getChainTypeValue(pathSpec) != 0) {
+			// in the future, other values might be supported
+			return PATH_INVALID;
+		}
+		if (isHardened(bip44_getAddressValue(pathSpec))) {
```
davidmisiak, line 434: Maybe we could check this also in `bip44_isGovernanceVotingKeyPath`?

src/securityPolicy.c
janmazak: Possibly, but it would not fit in a single screen, so I'd rather avoid that if no one is asking.

PR #183, opened 2022-11-30
Governance voting

janmazak:


gitmachtl:
I saw that the voting key on the previous ledger app 5.0.0 was displayed on the GUI as an ed25519 bech key. Was this changed to gov_vk now? Because thats the new standard for vote/governance-public keys.

PR #185, opened 2022-12-08
update the list of in-app tokens

janmazak:


janmazak:
Based on #184 .

janmazak:
Merged by Ledger in their repo.

PR #186, opened 2022-12-09
fix: validation of P2 values

janmazak:


refi93 (approved):

src/signTx.c
```diff
@@ -772,9 +772,10 @@ static void signTx_handleAuxDataAPDU(uint8_t p2, const uint8_t* wireDataBuffer,
 
 			signTxGovernanceVotingRegistration_handleAPDU(p2, wireDataBuffer, wireDataSize);
 			return;
-		} else {
-			CHECK_STAGE(SIGN_STAGE_AUX_DATA);
 		}
+
+		VALIDATE(p2 == P2_UNUSED, ERR_INVALID_REQUEST_PARAMETERS);
+		CHECK_STAGE(SIGN_STAGE_AUX_DATA);
```
refi93, line 778: seems legit, can you just clarify what issue did the previous code cause?

janmazak: There were new APDUs added to the state machine, and thus they were not recognized by old ledger app, and not catched by the above submachine `if`. Consequently, we did not do any check for P2 values (not belonging to the old submachine).

Now we first check P2 and then, based on valid P2, we decide whether we are in a submachine or not.

janmazak:
Merged by Ledger in their repo.

PR #192, opened 2023-06-05
Update tokenList.json (swapping MELD token old->new)

gitmachtl:
@janmazak 
- Swapping out the MELD token assetSubject from the old to the new one in the tokenList.json
- Fingerprint data in the token_data.c file also changed according to the new assetSubject


janmazak:
Will be resolved via
https://github.com/LedgerHQ/app-cardano/pull/44

PR #197, opened 2023-08-15
update crypto calls

janmazak:


janmazak:
See https://github.com/LedgerHQ/app-cardano/pull/53

PR #198, opened 2023-08-16
allow multidelegation keys

janmazak:
No difference in UI treatment between keys with index 0 and more than 0, as discussed with IOG and others.


janmazak:
https://github.com/LedgerHQ/app-cardano/pull/53 should be processed before this.

davidmisiak (approved):

PR #209, opened 2024-01-06
Conway tag258 + max URL length

janmazak:


davidmisiak (approved):
Apart from the comments, LGTM (but haven't tested).

janmazak:
included in #208

